PR #13, opened 2023-10-09
Fixed up some doc errors

packquickly:
Switched verbose from being a bool to frozenset of strings, and set `use_inverse=False` for the mix-and-match example (temporarily, I'll add proper support for `use_inverse=True` soon.)


patrick-kidger:
LGTM!

PR #34, opened 2024-01-06
Check last point when using best so far minimiser

### LINKED ISSUE #33: BestSoFarMinimiser behavior
Not sure if this is a bug or not, but `BestSoFarMinimiser` appears to not check the last step of the wrapped solver: 

```
solver = optimistix.BestSoFarMinimiser(optimistix.BFGS(rtol=1e-5, atol=1e-5))
ret = optimistix.minimise(lambda x, _: (x - 3.)**2, solver, 0.)
print(ret.value, ret.state.state.y_eval)

0.0 3.0
```

ColCarroll:
Fixes #33 

There is some unnecessary looking lines

```
        best_f, best_aux = fn(state.best_y, args)
        best_loss = self._to_loss(state.best_y, best_f)
```

where I would expect to just use `state.best_loss`, but the test doesn't pass without it!

### DIFF
--- a/optimistix/_solver/best_so_far.py
+++ b/optimistix/_solver/best_so_far.py
@@ -113,7 +113,16 @@ class _BestSoFarSolver(AbstractIterativeSolver, Generic[Y, Out, Aux]):
         tags: frozenset[object],
         result: RESULTS,
     ) -> tuple[Y, Aux, dict[str, Any]]:
-        return state.best_y, state.best_aux, {}
+        f, aux = fn(y, args)
+        loss = self._to_loss(y, f)
+
+        best_f, best_aux = fn(state.best_y, args)
+        best_loss = self._to_loss(state.best_y, best_f)
+
+        pred = loss < best_loss
+        best_y = tree_where(pred, y, state.best_y)
+        best_aux = tree_where(pred, aux, state.best_aux)
+        return best_y, best_aux, {}
 
 
 class BestSoFarMinimiser(  # pyright: ignore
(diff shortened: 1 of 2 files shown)


optimistix/_solver/best_so_far.py
```diff
@@ -113,7 +113,16 @@ def postprocess(
         tags: frozenset[object],
         result: RESULTS,
     ) -> tuple[Y, Aux, dict[str, Any]]:
-        return state.best_y, state.best_aux, {}
+        f, aux = fn(y, args)
+        loss = self._to_loss(y, f)
+
+        best_f, best_aux = fn(state.best_y, args)
+        best_loss = self._to_loss(state.best_y, best_f)
```
patrick-kidger, line 120: Agreed, I definitely don't understand these lines. What do you think is going on?

tests/test_best_so_far.py
```diff
@@ -41,3 +41,13 @@ def fn(y, _):
...
+
+
+def test_checks_last_point():
+    def fn(y, _):
+        return (y - 3.0) ** 2
+
+    solver = optx.BestSoFarMinimiser(optx.BFGS(rtol=1e-5, atol=1e-5))
+    sol = optx.minimise(fn, solver, jnp.array(0.0))
+    assert sol.value == 3.0
+    # assert fn(sol.value, None) <= fn(sol.state.state.y_eval, None)
```
patrick-kidger, line 53: Can we remove this line?

patrick-kidger:
Thank you for the fix! Always happy to squash bugs :)

Issue #50, opened 2024-03-17
Incompatibility of least_squares and custom_vjp


ahwillia:
I'm running into some trouble applying `optimistix.least_squares(fn, LevenbergMarquardt(...), x0)` to certain problems. From the error message below, my understanding of the root cause is that forward-mode autodiff cannot be used on `jax.custom_vjp`. In my case I am using `diffrax` to solve an ODE within `fn(...)`, which I think might be causing the problem. 

Is my basic understanding correct? Are there specific constraints / assumptions that `fn(...)` must follow for `optimistix.least_squares` to work (e.g. cannot use `jax.custom_vjp`)? Is there any way around this?

The error I get is:

```
TypeError: can't apply forward-mode autodiff (jvp) to a custom_vjp function.
```

The full code to reproduce the error is below. By the way I get the same problem when trying to use `jaxopt.LevenbergMarquardt` on this problem.

```python
# === imports === #
from jax.config import config; config.update("jax_enable_x64", True)
import jax
import jax.numpy as jnp
import diffrax
from diffrax import ODETerm, Dopri5, SaveAt
from tqdm import trange
import optimistix
from optimistix import LevenbergMarquardt

# === functions defining flow field and residuals === #
def geodesic_vector_field(P):
    jacP = jax.jacobian(P)
    def vector_field(t, state, args):
        x, v = state
        Pdx = jacP(x)
        q1 = 0.5 * jnp.einsum("jki,j,k->i",Pdx, v, v)
        q2 = jnp.einsum("ilp,l,p->i", Pdx, v, v)
        dxdt = v
        dvdt = jnp.linalg.solve(P(x), q1 + q2)
        return (dxdt, dvdt)
    return vector_field

def exponential_map(x0, v0, term, solver):
    return diffrax.diffeqsolve(
        term, solver, t0=0, t1=1, dt0=0.1, y0=(x0, v0),
        saveat=SaveAt(t0=False, t1=True)
    ).ys[0].ravel()

def shooting_method_resids(x0, x1, term, solver):
    return jax.jit(
        lambda v0, args: (x1 - exponential_map(x0, v0, term, solver)).ravel()
    )

# === try solving the boundary value problem === #
term = ODETerm(geodesic_vector_field(lambda x: jnp.eye(2)))
solver = Dopri5()

optimistix.least_squares(
    shooting_method_resids(jnp.zeros(2), jnp.ones(2), term, solver),
[...]

PR #51, opened 2024-03-18
AbstractGaussNewton now supports reverse-autodiff for Jacobians.

patrick-kidger:
In particular this is useful when the underlying function only supports reverse-mode autodifferentiation due to a `jax.custom_vjp`, see https://github.com/patrick-kidger/optimistix/issues/50

### DIFF
--- a/optimistix/_solver/dogleg.py
+++ b/optimistix/_solver/dogleg.py
@@ -224,6 +224,13 @@ class Dogleg(AbstractGaussNewton[Y, Out, Aux], strict=True):
     The distinguishing feature of this algorithm is the "dog leg" shape of its descent
     path, in which it begins by moving in the steepest descent direction, and then
     switches to moving in the Newton direction.
+
+    Supports the following `options`:
+
+    - `jac`: whether to use forward- or reverse-mode autodifferentiation to compute the
+        Jacobian. Can be either `"fwd"` or `"bwd"`. Defaults to `"fwd"`, which is
+        usually more efficient. Changing this can be useful when the target function has
+        a `jax.custom_vjp`, and so does not support forward-mode autodifferentiation.
     """
 
     rtol: float
--- a/optimistix/_solver/gauss_newton.py
+++ b/optimistix/_solver/gauss_newton.py
@@ -1,5 +1,5 @@
 from collections.abc import Callable
-from typing import Any, Generic, Optional, Union
+from typing import Any, Generic, Literal, Optional, Union
 
 import equinox as eqx
 import jax
@@ -164,10 +164,28 @@ class _GaussNewtonState(
 
 
 def _make_f_info(
-    fn: Callable[[Y, Args], tuple[Any, Aux]], y: Y, args: Args, tags: frozenset
+    fn: Callable[[Y, Args], tuple[Any, Aux]],
+    y: Y,
+    args: Args,
+    tags: frozenset,
+    jac: Literal["fwd", "bwd"],
 ) -> tuple[FunctionInfo.ResidualJac, Aux]:
-    f_eval, lin_fn, aux_eval = jax.linearize(lambda _y: fn(_y, args), y, has_aux=True)
-    jac_eval = lx.FunctionLinearOperator(lin_fn, jax.eval_shape(lambda: y), tags)
+    if jac == "fwd":
+        f_eval, lin_fn, aux_eval = jax.linearize(
+            lambda _y: fn(_y, args), y, has_aux=True
+        )
+        jac_eval = lx.FunctionLinearOperator(lin_fn, jax.eval_shape(lambda: y), tags)
+    elif jac == "bwd":
+        # Materialise the Jacobian in this case.
+        def _for_jacrev(_y):
+            f_eval, aux_eval = fn(_y, args)
+            return f_eval, (f_eval, aux_eval)
+
+        jac_pytree, (f_eval, aux_eval) = jax.jacrev(_for_jacrev, has_aux=True)(y)
+        output_structure = jax.eval_shape(lambda: f_eval)
+        jac_eval = lx.PyTreeLinearOperator(jac_pytree, output_structure, tags)
+    else:
+        raise ValueError("Only `jac='fwd'` or `jac='bwd'` are valid.")
     return FunctionInfo.ResidualJac(f_eval, jac_eval), aux_eval
 
 
@@ -187,6 +205,13 @@ class AbstractGaussNewton(
     - `descent`: `AbstractDescent`
     - `search`: `AbstractSearch`
     - `verbose`: `frozenset[str]`
+
+    Supports the following `options`:
+
+    - `jac`: whether to use forward- or reverse-mode autodifferentiation to compute the
+        Jacobian. Can be either `"fwd"` or `"bwd"`. Defaults to `"fwd"`, which is
+        usually more efficient. Changing this can be useful when the target function has
+        a `jax.custom_vjp`, and so does not support forward-mode autodifferentiation.
     """
 
     rtol: AbstractVar[float]
@@ -208,7 +233,8 @@ class AbstractGaussNewton(
         aux_struct: PyTree[jax.ShapeDtypeStruct],
         tags: frozenset[object],
     ) -> _GaussNewtonState:
-        f_info_struct, _ = eqx.filter_eval_shape(_make_f_info, fn, y, args, tags)
+        jac = options.get("jac", "fwd")
+        f_info_struct, _ = eqx.filter_eval_shape(_make_f_info, fn, y, args, tags, jac)
         f_info = tree_full_like(f_info_struct, 0, allow_static=True)
         return _GaussNewtonState(
             first_step=jnp.array(True),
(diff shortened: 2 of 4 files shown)


johannahaffner:
Hi Patrick, 

I tried this branch, and it did not work for my use case (parameter estimation for an ODE). I tried both `LevenbergMarquardt` and `GaussNewton` in combination with `DirectAdjoint` or `RecursiveCheckpointAdjoint`, and I get  

```
TypeError: can't apply forward-mode autodiff (jvp) to a custom_vjp function.
```
with both. I have an MWE if that would be useful for you. 

This is the commit I installed from: 

```
Switched to a new branch 'gauss-newton-jacrev'
branch 'gauss-newton-jacrev' set up to track 'origin/gauss-newton-jacrev'.
Resolved https://github.com/patrick-kidger/optimistix to commit 776820485fd9df320d3089bcd302f2f69124cf14
```

Hope you had a nice weekend!

Johanna

johannahaffner:
Nevermind, it works! I checked out the other issue and realized that I need to pass `options=dict(jac="bwd")` to `least_squares`, now it works.

patrick-kidger:
Awesome, I'm glad to hear it! I've just merged this in, so this will appear in the next release of Optimistix. :)

johannahaffner:
Hi! Not sure where else to post this, but I wanted to note it somewhere: on my problem, I get dramatic performance reductions when using reverse-mode autodiff inside of `least_squares`. 

Specifically: 

- Using `dfx.DirectAdjoint`, I get 0.286 s per trajectory if passing `options=dict(jac="fwd")` and
- 29.8 s per trajectory if passing `options=dict(jac="bwd")`, which drops to
- 8.11 s per trajectory if using `dfx.RecursiveCheckpointAdjoint` instead.

The documentation does note that forward-mode is usually more efficient, so this might just be further confirmation of that observation :)

patrick-kidger:
Hey there!

Ah, indeed this might have been the case. It was hard to guess which was the greater overhead: `DirectAdjoint + forward mode`, or `RecursiveCheckpointAdjoint + reverse mode`.

FWIW this is motivating me to consider adding a forward-mode specific "adjoint". (I'd been holding off on this in the hopes that support could be added to `RecursiveCheckpointAdjoint` but that's something which requires changes in JAX itself.) Let me see if I can throw something together and we can see how it runs.

patrick-kidger:
Okay, completely untested, but something like the following should probably work:
```python
class ForwardMode(AbstractAdjoint):
    def loop(
        self,
        *,
        solver,
        throw,
        passed_solver_state,
        passed_controller_state,
        **kwargs,
    ):
        del throw, passed_solver_state, passed_controller_state
        inner_while_loop = ft.partial(_inner_loop, kind="lax")
        outer_while_loop = ft.partial(_outer_loop, kind="lax")
        # Support forward-mode autodiff.
        # TODO: remove this hack once we can JVP through custom_vjps.
        if isinstance(solver, AbstractRungeKutta) and solver.scan_kind is None:
            solver = eqx.tree_at(
                lambda s: s.scan_kind, solver, "lax", is_leaf=_is_none
            )
        final_state = self._loop(
            solver=solver,
            inner_while_loop=inner_while_loop,
            outer_while_loop=outer_while_loop,
            **kwargs,
        )
        return final_state
```
by passing `diffeqsolve(..., adjoint=ForwardMode())`. Then the resulting `diffeqsolve` should be forward-mode autodifferentiable.

johannahaffner:
Awesome, thank you! I'm happy to give it a proper go tomorrow. So far I am getting an error in diffrax 

```console
.../site-packages/diffrax/_integrate.py#line=457), in loop()
    456     return new_state
--> 458 final_state = outer_while_loop(
    459     cond_fun, body_fun, init_state, max_steps=max_steps, buffers=_outer_buffers
    460 )
    462 def _save_t1(subsaveat, save_state):

.../contextlib.py#line=80), in inner()
     80 with self._recreate_cm():
---> 81     return func(*args, **kwds)

.../site-packages/equinox/internal/_loop/loop.py#line=102), in while_loop()
    102 del cond_fun, body_fun, init_val
--> 103 _, _, _, final_val = lax.while_loop(cond_fun_, body_fun_, init_val_)
    104 return final_val

JaxStackTraceBeforeTransformation: ValueError: Reverse-mode differentiation does not work for 
lax.while_loop or lax.fori_loop with dynamic start/stop values. Try using lax.scan, or using 
fori_loop with static start/stop.
```

I'm a little unsure why diffrax is attempting reverse-mode differentiation here. But also happy to dig into it a little :)

johannahaffner:
Ok, mini update: I haven't been able to dig into it any further, but I did get better benchmarks on my problem.
I use `jit(vmap(...))` to parallelize over the individuals, so the runtime is determined by the trajectories that take longest to fit.
The sum of the maximum number of steps taken in each iteration works out to a value that is quite close to the total runtime, divided by the time it takes to simulate the data. (376 vs. 380.)

It looks like forward-mode differentiation on `DirectAdjoint` is already super fast, at least as far as `optimistix` is concerned. So unless the `ForwardModeAdjoint` also speeds up the ODE solving, I would not expect too many performance gains to come from this.

patrick-kidger:
> So unless the ForwardModeAdjoint also speeds up the ODE solving

Indeed it does! (Granted, possibly not by that much. Depends on your problem.)

Do you have a quick MWE demonstrating the above crash?

johannahaffner:
Ah even faster ODEs! Nice, that would be cool. 
I tried this on my real data, will make an MWE after the weekend!

johannahaffner:
Here comes the MWE! The adjoint works fine in `diffrax`. But optimistix attempts to reverse-mode differentiate through it.

```python
import jax.numpy as jnp

import equinox as eqx
import diffrax
import optimistix

import functools

from jax import Array

_is_none = lambda x: x is None  # This function is needed in the forward adjoint

class ForwardMode(diffrax.AbstractAdjoint):
    def loop(
        self,
        *,
        solver,
        throw,
        passed_solver_state,
        passed_controller_state,
        **kwargs,
    ):
        del throw, passed_solver_state, passed_controller_state
        inner_while_loop = functools.partial(diffrax._adjoint._inner_loop, kind="lax")
        outer_while_loop = functools.partial(diffrax._adjoint._outer_loop, kind="lax") 
        # Support forward-mode autodiff.
        # TODO: remove this hack once we can JVP through custom_vjps.
        if isinstance(solver, diffrax.AbstractRungeKutta) and solver.scan_kind is None:
            solver = eqx.tree_at(
                lambda s: s.scan_kind, solver, "lax", is_leaf=_is_none
            )
        final_state = self._loop(
            solver=solver,
            inner_while_loop=inner_while_loop,
            outer_while_loop=outer_while_loop,
            **kwargs,
        )
        return final_state

class ToyModel(eqx.Module):
    """Toy model that provides a simple interface to generate an ODE solution,
    subject to its parameters.
    """
    _term: diffrax.ODETerm
    _t0: float
    _t1: float
    _dt0: float
    _y0: float
    _saveat: diffrax.SaveAt
    _solver: diffrax.AbstractERK
    _adjoint: diffrax.AbstractAdjoint

    def __init__(self, ode_model, initial_state, times, solver, adjoint):
        self._term = diffrax.ODETerm(ode_model)
        self._y0 = initial_state
        
        self._t0 = times[0]
        self._t1 = times[-1]
        self._dt0 = 0.01
        self._saveat = diffrax.SaveAt(ts=times)
        
        self._solver = solver
        self._adjoint = adjoint

    def __call__(self, param):
        sol = diffrax.diffeqsolve(
            self._term, 
            self._solver, 
            self._t0, self._t1, self._dt0, self._y0, 
            args=param, 
            saveat=self._saveat, 
            adjoint=self._adjoint,
        )
        return sol.ys

def estimate_parameters(initial_guess, model, data, solver, solver_options: dict = dict(jac='fwd')):
    """Function that estimates the parameters."""

    def residuals(param, args):
        model, data = args
        fit = model(param)
        res = data - fit
        return res

    sol = optimistix.least_squares(
        residuals, 
        solver, 
        initial_guess,
        args = (model, data),
        options = solver_options,
    )
    return sol

# Create the model
def dydt(t, y, k):  # Toy ODE
    return - k * y
t = jnp.linspace(0, 10, 50)  
y0 = 10.
model = ToyModel(dydt, y0, t, diffrax.Tsit5(), ForwardMode())

# Solve ODE
k = 0.5  # True value
ode_solution = model(k)  # This runs without issue

# Now try solving for the parameters
k0 = 0.1
solver = optimistix.LevenbergMarquardt(atol=1e-09, rtol=1e-06)
lm_solution = estimate_parameters(k0, model, ode_solution, solver)  # This fails
```

patrick-kidger:
Can you try https://github.com/patrick-kidger/optimistix/pull/61? Hopefully that should fix things :)

(It's also revealed an unrelated bug with complex numbers, but I don't think that should affect you.)

johannahaffner:
Hi Patrick, 

amazing, thank you for the quick fix! 
I tried it on my real data, it works a charm and delivers a handsome 3x speedup.

johannahaffner:
> delivers a handsome 3x speedup.

whoops, my bad, I forgot to control for whether my laptop was plugged in :) 

Using `DirectAdjoint` is 52 % slower than using `ForwardMode` for the parameter estimation, and solving just the ODE is 42 % slower with `DirectAdjoint. 

Still pretty nice! I'll keep using the ForwardMode above.

PR #68, opened 2024-07-13
Add transpose rule for stop_gradient

dkweiss31:
Work around for JAX issue #22011, as well as https://github.com/patrick-kidger/diffrax/pull/387#issuecomment-2174488365


patrick-kidger:
So in direct contradiction to my past self...

https://github.com/patrick-kidger/diffrax/pull/387#issuecomment-2199206730

dkweiss31:
Yes of course!

PR #69, opened 2024-07-16
Clarify task of descent classes

johannahaffner:
Nit: I find "size of the update" to be confusing in this sentence. As I understand it, the descent computes the actual update, using the scalar computed in the search. The latter might be described more accurately as the "size" of the update.


patrick-kidger:
This LGTM! Thank you for the tidy-up :)

johannahaffner:
Thanks!

PR #77, opened 2024-08-17
Pass parameter value to optax's optimizer update

NeilGirdhar:
This is an optional parameter that it doesn't hurt to pass.


patrick-kidger:
LGTM! Merged.
Thank you for your first PR!

PR #78, opened 2024-08-18
Allow the bisection root finder to expand the interval

NeilGirdhar:


optimistix/_solver/bisection.py
```diff
@@ -18,6 +20,61 @@ class _BisectionState(eqx.Module, strict=True):
...
+                          fn=fn,
+                          args=args)
+    body_fun = ft.partial(_expand_interval, expand_upper=expand_upper)
+    return eqxi.while_loop(cond_fun, body_fun, initial_interval, kind="lax")
+
+
+def _retain_interval(initial_interval: _ExpansionCarry,
+                     upper_val: Scalar,
+                     lower_val: Scalar,
+                     need_positive: Bool[Array, ""],
```
patrick-kidger, line 73: Nit: I think we can simplify the `cond` by partially-applying these arguments to the `_expand_interval_repeatedly` branch, so that only the `_ExpansionCarry` object is passed to `lax.cond(..., initial_interval)`.

optimistix/_solver/bisection.py
```diff
@@ -18,6 +20,61 @@ class _BisectionState(eqx.Module, strict=True):
...
+                                args: PyTree,
+                                ) -> _ExpansionCarry:
+    expand_upper = need_positive ^ (upper_val < lower_val)
+    cond_fun = ft.partial(_interval_contains_root,
+                          need_positive=need_positive,
+                          expand_upper=expand_upper,
+                          fn=fn,
+                          args=args)
+    body_fun = ft.partial(_expand_interval, expand_upper=expand_upper)
+    return eqxi.while_loop(cond_fun, body_fun, initial_interval, kind="lax")
```
patrick-kidger, line 67: If it's going to be marked with `kind="lax"` then I'd suggest just using `lax.while_loop` directly.

(For context, the main purpose of `eqxi.while_loop` is to be a backpropagatable while loop, which is a thing accomplished using checkpointing. We have two main implementations, `kind="checkpointed"` -- the only to use 99% of the time -- and `kind="bounded"` -- which is an older first attempt. For completeness I also added the ability to fall back to just `kind="lax"`, but that's only really useful if you want to be polymorphic over loop implementations.)

optimistix/_solver/bisection.py
```diff
@@ -74,11 +136,26 @@ def init(
...
-                lower_neg == upper_neg,
-                msg="The root is not contained in [lower, upper]",
-            )
+            root_not_contained = lower_neg == upper_neg
+            if self.expand_if_necessary:
+                initial_interval = _ExpansionCarry(lower, upper)
+                need_positive = lower_neg
+                final_interval = jax.lax.cond(root_not_contained,
+                                              ft.partial(_expand_interval_repeatedly,
+                                                         fn=fn, args=args),
```
patrick-kidger, line 145: Optionally, by the way, I think we can improve compile time here rather a lot: I suspect it should be possible to fold the two `fn` evaluations above into this loop. We just always run at least one (or two?) iterations, equivalent to the two evaluations we're currently doing.

This would both remove two callsites of `fn` and remove the surrounding `cond`.

I can see that might be a fiddlier change, though, so I'm not asking for that as a requirement to get this merged.

patrick-kidger (commented):
This LGTM! Would you be able to add some tests for the new functionality as well, please?

NeilGirdhar:
> This LGTM! Would you be able to add some tests for the new functionality as well, please?

Of course, just wanted to make sure the interface was okay with you first 😄

optimistix/_solver/bisection.py
NeilGirdhar: Should we switch to checkpointed?  Will this affect our ability to take derivatives through this computation?

patrick-kidger: If we do this right, then I think this shouldn't affect our ability to compute derivatives. The mathematical definition of the gradient shouldn't be affected by our choice of interval.

patrick-kidger:
Haha! Great stuff. Definitely happy with how this looks :)

optimistix/_solver/bisection.py
NeilGirdhar: Sounds great.

optimistix/_solver/bisection.py
NeilGirdhar: I was able to remove the `cond` as requested (it's actually how I originally had it), but I can't see how to fold the functions in.

optimistix/_solver/bisection.py
NeilGirdhar: Good tip for the future, but now that I've taken your other suggestion, this isn't relevant anymore.

NeilGirdhar:
I added some simple tests.

optimistix/_solver/bisection.py
```diff
@@ -18,6 +19,63 @@ class _BisectionState(eqx.Module, strict=True):
...
+    return need_positive ^ (carry_val > 0.0)
+
+
+def _expand_interval(
+    carry: _ExpansionCarry,
+    /,
+    *,
+    expand_upper: Bool[Array, ""],
+) -> _ExpansionCarry:
+    new_domain = 2.0 * (carry.upper - carry.lower)
```
patrick-kidger, line 47: Why `2 * `? This means we triple the length of the interval each time?

patrick-kidger (commented):
Question aside, this LGTM! Let me know what you think and then I'll merge this.

optimistix/_solver/bisection.py
NeilGirdhar: Sorry if this was confusing, but I shift lower and upper while doubling the interval.

For example:
lower = 1
upper = 4
f(lower) = -10
f(upper) = -5

So, we want to move right.  We know the root is not in [1, 4], and we want to double the interval length.  Thus:

new_lower = 4
new_upper = 2 * 3 + 4 = 10

What do you think?

patrick-kidger: Ah, understood! LGTM :)

PR #80, opened 2024-08-23
Add py.type as per PEP 561

NeilGirdhar:
You have great type annotations, but they're invisible to callers!


patrick-kidger:
You're right!
Merged! :D

PR #85, opened 2024-09-27
Remove unused inner function

johannahaffner:
A little thing I came across while slowly chipping away at my SQP implementation.

This inner function does not seem to be used anywhere, and also does not seem to be an intentional holdover, as far as I can tell.


patrick-kidger:
LGTM! Looks like the failing test is due to a breaking change in JAX that affects Lineax. I've opened https://github.com/patrick-kidger/lineax/pull/110 as a possible fix (+Lineax version bump). I'll ignore it for now -- merged! :)

PR #94, opened 2024-11-17
Small bug in BFGS - grad now computed with respect to current iterate in accepted branch

johannahaffner:
Hi Patrick, 

I think I spotted a small bug in BFGS. 

In the accepted branch, `lin_fn` is passed to `lin_to_grad` with input `y`, which is the last accepted iterate _before_ we decided to accept the current iterate `y_eval`. 
We have linearised `f` at `y_eval`, but are now evaluating this at `y`. 

If I'm not mistaken, then this means that our BFGS update formula implements 

$\nabla f(y_{k}) - \nabla f(y_{k-1})$ instead of $\nabla f(y_{k + 1}) - \nabla f(y_{k})$, where `y_eval` corresponds to $y_{k+1}$.

PS: feel free to disregard the .venv in .gitignore, doesn't need to be in main :)

### DIFF
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -119,9 +119,8 @@ def jacobian(fn, in_size, out_size, has_aux=False):
 
 
 def lin_to_grad(lin_fn, *primals):
-    # Only the shape and dtype of primals is evaluated, not the value itself. lin_fn
-    # already knows where it was computed, and 1.0 acts only as a scaling factor here.
-    # We compute lin_fn and grad separately because it helps us compile efficiently, see
+    # Only the shape and dtype of primals is evaluated, not the value itself. We convert
+    # to grad after linearising to avoid recompilation.
     # https://github.com/patrick-kidger/optimistix/issues/89#issuecomment-2447669714
     return jax.linear_transpose(lin_fn, *primals)(1.0)
 
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -120,7 +120,7 @@ def jacobian(fn, in_size, out_size, has_aux=False):
 
 def lin_to_grad(lin_fn, *primals):
     # Only the shape and dtype of primals is evaluated, not the value itself. We convert
-    # to grad after linearising to avoid recompilation.
+    # to grad after linearising to avoid recompilation. (1.0 is a scaling factor.)
     # https://github.com/patrick-kidger/optimistix/issues/89#issuecomment-2447669714
     return jax.linear_transpose(lin_fn, *primals)(1.0)
 
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -119,6 +119,10 @@ def jacobian(fn, in_size, out_size, has_aux=False):
 
 
 def lin_to_grad(lin_fn, *primals):
+    # Only the shape and dtype of primals is evaluated, not the value itself. lin_fn
+    # already knows where it was computed, and 1.0 acts only as a scaling factor here.
+    # We compute lin_fn and grad separately because it helps us compile efficiently, see
+    # https://github.com/patrick-kidger/optimistix/issues/89#issuecomment-2447669714
     return jax.linear_transpose(lin_fn, *primals)(1.0)
 
 
--- a/optimistix/_solver/bfgs.py
+++ b/optimistix/_solver/bfgs.py
@@ -216,7 +216,7 @@ class AbstractBFGS(
         )
 
         def accepted(descent_state):
-            (grad,) = lin_to_grad(lin_fn, y)
+            (grad,) = lin_to_grad(lin_fn, state.y_eval)
             y_diff = (state.y_eval**ω - y**ω).ω
             if self.use_inverse:
                 hessian = None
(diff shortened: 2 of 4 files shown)


patrick-kidger:
Ooh, that's quite concerning!

So I think the same bug also exists in `AbstractGradientDescent`.

Just reading the code, I think you're correct. Did you come across this as a result of a performance drop or something? (I.e. do we know how much this may have affected things?)

johannahaffner:
I haven't measured this, no. It just confused me while implementing other things. 

I'll run this interactively and record some traces!

johannahaffner:
...I'm not seeing *any* differences. I tried this on Rosenbrock, for $a = 1$ and $b=100$, with a few different tolerances and two different initial points. I focused more on BFGS, but I ran GradientDescent as well.

Now I am really confused :D And I think even though

- Initial approximate Hessians may not be great (BFGS)
- the line search might help
- closer to the optimum it should be after to rely on slightly outdated information 

This should make at least a little bit of a difference. I'll take a look again tomorrow - do you have another benchmarking problem in mind?

patrick-kidger:
Haha! That's very weird. So given that this is based around the accepted part of a search, then I'm guessing the choice of search here probably affects things quite a lot? I'd play around with that.

And how could I possibly have a benchmarking problem I prefer more than Rosenbrock? That would be like if there existed a food you like more than ice cream.

(More seriously, we have a collection of benchmark problems in `tests/helpers.py`.)

johannahaffner:
Comparing among the solvers helped! I realised that the value of the gradient in the accepted branch does not depend on the input value passed to `lin_to_grad`. This is because it uses `jax.linear_transpose` under the hood

https://github.com/patrick-kidger/optimistix/blob/ef86ef5d3dcd1d4be61611137157d19bf7944118/optimistix/_misc.py#L122

which will only evaluate the shape and dtype of `*primals`, and not use its values. (The 1.0 is basically just a scaling factor.)
I suggest we still call `lin_to_grad` with `y_eval`, and optionally also document that it acts as a dummy value here.

> And how could I possibly have a benchmarking problem I prefer more than Rosenbrock? That would be like if there existed a food you liked more than ice cream!

Raspberries, frozen, straight from the package. It's like crunchy ice cream :D

johannahaffner:
Alternatively we could make this even more explicit by passing `tree_full_like(y, 0)` in the accepted branch.

optimistix/_misc.py
```diff
@@ -119,9 +119,8 @@ def jacobian(fn, in_size, out_size, has_aux=False):
 
 
 def lin_to_grad(lin_fn, *primals):
-    # Only the shape and dtype of primals is evaluated, not the value itself. lin_fn
-    # already knows where it was computed, and 1.0 acts only as a scaling factor here.
-    # We compute lin_fn and grad separately because it helps us compile efficiently, see
+    # Only the shape and dtype of primals is evaluated, not the value itself. We convert
+    # to grad after linearising to avoid recompilation.
```
patrick-kidger, line 123: I think the commentary on `1.0` is still useful?

patrick-kidger:
> I realised that the value of the gradient in the accepted branch does not depend on the input value passed to lin_to_grad. This is because it uses jax.linear_transpose under the hood

Aha! Mystery solved. :D

> I suggest we still call lin_to_grad with y_eval, and optionally also document that it acts as a dummy value here.

Agreed.

> Raspberries, frozen, straight from the package. It's like crunchy ice cream :D

Pffft, the world takes all sorts I suppose. ;)

optimistix/_misc.py
johannahaffner: Alright, then let's keep it!

johannahaffner:
Haha yes, and such a simple thing in the end :D but future me will be happy about the comments.

> Pffft, the world takes all sorts I suppose. ;)
Give them a try! 

...now that I'm typing this: is the failed check due to the new jaxtyping?

optimistix/_solver/bfgs.py
```diff
@@ -216,7 +216,7 @@ def step(
         )
 
         def accepted(descent_state):
-            (grad,) = lin_to_grad(lin_fn, y)
+            (grad,) = lin_to_grad(lin_fn, tree_full_like(y, 0))
```
patrick-kidger, line 219: I feel like passing zeros here is equally confusing as your first point, to be honest. I'd probably either just make this `state.y_eval` and/or add a comment that the value here doesn't matter.

patrick-kidger:
Yes it is :(
I'll do a new jaxtyping release with the fix.

optimistix/_solver/bfgs.py
johannahaffner: Alright! I now do both.

patrick-kidger: Wonderful stuff! This LGTM.

patrick-kidger:
And merged! I'm ignoring the jaxtyping failure, which is fixed in https://github.com/patrick-kidger/jaxtyping/pull/271.

johannahaffner:
Thanks! Figuring this out was fun :)

Issue #56, opened 2024-05-07
Using `optimistix` with an `equinox` model


frostedoyster:
Hi everyone, thanks for the great library and apologies in advance for this basic question.
I'm trying to find the true minimum of a small neural network, and I thought of using a solver from `optimistix` together with an `equinox` model. However, I haven't been able to make the two work together.

Here is a minimal snippet which fails:

```
import jax 
import jax.numpy as jnp
import equinox as eqx
import optimistix as optx

jax.config.update("jax_enable_x64", True)


X = jax.random.normal(jax.random.PRNGKey(0), (2000, 8))

@jax.vmap
def function(x):
    return x[0] + x[1]**2 + jnp.cos(x[2]) + jnp.sin(x[3]) + x[4]*x[5] + (x[6]*x[7])**3

y = function(X).reshape(-1, 1)

model = eqx.nn.MLP(in_size=8, out_size=1, width_size=4, depth=2, activation=jax.nn.silu, key=jax.random.PRNGKey(0))

static, params = eqx.partition(model, eqx.is_inexact_array)

def loss_fn(params, static, X, y):
    model = eqx.combine(params, static)
    return jnp.sum((jax.vmap(model)(X) - y)**2)

solver = optx.Newton(rtol=1e-5, atol=1e-5)
sol = optx.minimise(loss_fn, solver, params)
```

I'm getting `TypeError: Cannot determine dtype of <PjitFunction of <function silu at 0x742fde959300>>`.

What am I doing wrong?
Thank you in advance.

patrick-kidger:
You have `static` and `params` the wrong way around. 

(Caveat: I've not tried running the code, this was just what jumped out at me.)

frostedoyster:
Indeed, that was the issue! Thanks a lot for the reply and for the library!

raj-brown:
@patrick-kidger I ran the same code after fixing the param static but I get following errors
```
 File "/Users/raj/opt/anaconda3/envs/equinox/lib/python3.9/site-packages/optimistix/_misc.py", line 89, in __call__
    return self.fn(*args, **kwargs), None
  File "/Users/raj/WORK_RAJ/NEURONSE_DARPA/RANS_FSBL/CODES/test.py", line 24, in loss_fn
    model = eqx.combine(params, static)
  File "/Users/raj/opt/anaconda3/envs/equinox/lib/python3.9/site-packages/equinox/_filters.py", line 209, in combine
[...]

PR #95, opened 2024-12-07
Add a verbose option to BFGS. 

johannahaffner:
Adds a verbose option to BFGS, with supports printing out the values of `loss`, `y`, and `step_size` at every step. 
The tested custom solvers are updated, and the documentation is updated as well. 

Users that have written custom solvers that subclass BFGS will now see `ValueError: The following fields were not initialised during __init__: {'verbose'}` upon upgrading. 

I think this - plus a note in the next release - is probably enough for the subset of users who define custom solvers (and probably know optimistix quite well). It is raised from inside equinox, though, which makes it a little less obvious. I'm happy to improve this - WDYT?

I have "tested" this by using it in a notebook, but not otherwise.

This follows up on https://github.com/patrick-kidger/optimistix/issues/56#issuecomment-2525092603.

### DIFF
--- a/optimistix/_solver/bfgs.py
+++ b/optimistix/_solver/bfgs.py
@@ -19,6 +19,7 @@ from .._misc import (
     max_norm,
     tree_dot,
     tree_full_like,
+    verbose_print,
 )
 from .._search import (
     AbstractDescent,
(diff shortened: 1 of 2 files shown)


patrick-kidger:
LGTM! Merged on to the `dev` branch :)

johannahaffner:
Thanks!

PR #110, opened 2025-01-17
Tiny fix: require Python>=3.10, since equinox now does

johannahaffner:
...and I think that means that everything built on top of equinox should too?

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -21,7 +21,7 @@ jobs:
   run-test:
     strategy:
       matrix:
-        python-version: [ 3.10, 3.11 ]
+        python-version: [ "3.10", "3.11" ]
         os: [ ubuntu-latest ]
       fail-fast: false
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


patrick-kidger:
Yup, LGTM! I think the CI will need updating in a similar fashion, though. (LMK if you need pointing at the relevant bits.)

johannahaffner:
> (LMK if you need pointing at the relevant bits.)

I'll try to figure it out myself first, that way I'll remember it better!

johannahaffner:
...but why does it try to run the tests in Python 3.1? :D

johannahaffner:
Not sure where that last test is defined? (The one that is not triggered by a pull request.)

Also, should we do 3.10 and 3.12?

.github/workflows/run_tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   run-test:
     strategy:
       matrix:
-        python-version: [ 3.9, 3.11 ]
+        python-version: [ "3.10", "3.11" ]
```
patrick-kidger, line 24: Well found! Including the need for quote marks :)

patrick-kidger (approved):

patrick-kidger:
Yeah, I think 3.10 and 3.12 would be better!

The 3.9 one is defined through the GitHub UI as a mandatory check that must pass. I'll delete it.

Issue #112, opened 2025-01-24
`minimize` and forward AD


vadmbertr:
Hi!

I'm facing a similar use-case as the one described here #50 but I would like to optimize using `minimize` rather than `least_squares`.
Is there any plan/option for supporting a solution similar to `options={"jac": "fwd"}`?

Thanks a lot!
Vadim

johannahaffner:
Hi Vadim, 

selection of forward- vs. reverse-mode autodiff is currently implemented at the solver level. Which solvers are you interested in using with reverse-mode autodiff? 

If you're using diffrax underneath: diffrax now has efficient forward-mode autodiff as well, with `diffrax.ForwardMode`.

vadmbertr:
Hi @johannahaffner,

Thanks for you reply!

I was interested in BFGS, but in forward mode. I was able to successfully solve a "least_squares" problem using forward mode and GaussNewton.
Indeed I'm using diffrax with diffrax.ForwardMode!

johannahaffner:
You're welcome! 

Could you share the error you are getting? `BFGS` already uses `jax.linearize`, which implements forward-mode automatic differentiation. So this should actually work out of the box.

vadmbertr:
Hi!

Here is a MWE to reproduce:

```python
from diffrax import diffeqsolve, Euler, ForwardMode, ODETerm, SaveAt
import jax.numpy as jnp
import optimistix as optx


def fn(y0, _=None):
    vector_field = lambda t, y, args: -y
    term = ODETerm(vector_field)
    solver = Euler()
    saveat = SaveAt(ts=[0., 1., 2., 3.])

    sol = diffeqsolve(term, solver, t0=0, t1=3, dt0=0.1, y0=y0, saveat=saveat, adjoint=ForwardMode())
    
    return sol.ys


def least_square(y0, _=None):
    ys = fn(y0)
    return jnp.sum(ys**2)


ls_sol = optx.least_squares(fn, optx.GaussNewton(rtol=1e-8, atol=1e-8), jnp.asarray(1.))
print(ls_sol.value)  # 0.0
min_sol = optx.minimise(least_square, optx.BFGS(rtol=1e-8, atol=1e-8), jnp.asarray(1.))  # error raised here
print(min_sol.value)
```

and I get the following:
`ValueError: Reverse-mode differentiation does not work for lax.while_loop or lax.fori_loop with dynamic start/stop values. Try using lax.scan, or using fori_loop with static start/stop.`

[...]

PR #114, opened 2025-02-03
Adds an option to support forward-mode automatic differentiation in all minimisers

johannahaffner:
Minimisers now accept a `"fwd"` and a `"bwd"` option, and `lin_to_grad` will use transposition to get the gradient with reverse-mode, while calling a custom `_jacfwd` that evaluates the linearised function with unit-pytrees when using forward-mode. Where relevant, gradient-computation related documentation has been moved to these two functions.

I have parameterised the tests for `minimise` and `minimise_jvp` to try both options, let me know if there are other tests I should also parameterise. I'm assuming everything else will also work if these two pass.

I have also removed a function `jacobian` from the `_misc` module, that used size-based heuristics to toggle between forward- and reverse-mode automatic differentiation. Since we now support explicit options in all solvers that compute derivatives, this function no longer fits our current way of doing things. (I don't think it was used anywhere anyway.)

### DIFF
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -8,6 +8,7 @@ import jax.extend as jex
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.tree_util as jtu
+import numpy as np
 from equinox.internal import ω
 from jaxtyping import Array, ArrayLike, Bool, PyTree, Scalar
 from lineax.internal import (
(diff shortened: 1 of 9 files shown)


johannahaffner:
This addresses https://github.com/patrick-kidger/optimistix/issues/112#issuecomment-2630242596.

optimistix/_solver/gradient_methods.py
```diff
@@ -120,6 +120,13 @@ class AbstractGradientDescent(
     - `norm: Callable[[PyTree], Scalar]`
     - `descent: AbstractDescent`
     - `search: AbstractSearch`
+
+    Supports the following `options`:
+
+    - `mode`: whether to use forward- or reverse-mode autodifferentiation to compute the
```
patrick-kidger, line 126: Do we use `mode` elsewhere? This isn't a word that obviously describes the kind of autodiff.

optimistix/_misc.py
```diff
@@ -103,26 +104,46 @@ def __call__(self, *args, **kwargs):
         return out, aux
 
 
-def jacobian(fn, in_size, out_size, has_aux=False):
-    """Compute the Jacobian of a function using forward or backward mode AD.
-
-    `jacobian` chooses between forward and backwards autodiff depending on the input
-    and output dimension of `fn`, as specified in `in_size` and `out_size`.
+def _jacfwd(lin_fn, pytree):
```
patrick-kidger, line 107: Is there a reason not to use the normal `jax.jacfwd`? I realize this is a linear function but it should still give the correct answer. And at least in the case I don't see any obvious inefficiency.

johannahaffner: Because compilation times are substantially worse even for trivial functions. Runtimes can probably still be optimised, though - see below.

Should I check if they are ~20 ms apart, or 100x apart for a more complicated optimisation variable and target function? It is still a little unintuitive to me what HLO does when analysing computation graphs.

```python
import jax
import jax.numpy as jnp
import optimistix as optx


def l2(y):
    return jnp.sum(y**2)


y0 = jnp.ones(3)
f_eval, lin_fn = jax.linearize(l2, y0)

_jacfwd_compiled = jax.jit(optx._misc._jacfwd)
_ = _jacfwd_compiled(lin_fn, y0)

jacfwd_compiled = jax.jit(jax.jacfwd(l2))
_ = jacfwd_compiled(y0)


print("Runtimes ---")
print("\nRuntime of _jacfwd on lin_fn:")
%timeit -r7 -n100 optx._misc._jacfwd(lin_fn, y0).block_until_ready()
print("\nRuntime of jax.jacfwd on lin_fn:")
%timeit -r7 -n100 jax.jacfwd(lin_fn)(y0).block_until_ready()
print("\nRuntime of jax.jacfwd on l2:")
%timeit -r7 -n100 jax.jacfwd(l2)(y0).block_until_ready()

print("\n\nCompilaton times ---")
print("\nCompilation time of _jacfwd on lin_fn:")
%timeit -r7 -n100 jax.jit(optx._misc._jacfwd).lower(lin_fn, y0).compile()
print("\nCompilation time of jax.jacfwd on lin_fn:")
%timeit -r7 -n100 jax.jit(jax.jacfwd(lin_fn)).lower(y0).compile()
print("\nCompilation time of jax.jacfwd on l2:")
%timeit -r7 -n100 jax.jit(jax.jacfwd(l2)).lower(y0).compile()
```

The output I get is

```
Runtimes ---

Runtime of _jacfwd on lin_fn:
1.45 ms ± 1.04 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

Runtime of jax.jacfwd on lin_fn:
964 μs ± 551 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

Runtime of jax.jacfwd on l2:
822 μs ± 253 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)


Compilaton times ---

Compilation time of _jacfwd on lin_fn:
198 μs ± 73.2 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

Compilation time of jax.jacfwd on lin_fn:
18.4 ms ± 291 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

Compilation time of jax.jacfwd on l2:
19 ms ± 243 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

optimistix/_solver/gradient_methods.py
johannahaffner: We use `jac` in the Gauss-Newton family of solvers.

optimistix/_misc.py
johannahaffner: PS: I get the same result when lowering a partially bound version of `_jacfwd`, such that the only argument of `lower` is `y0`.

johannahaffner: I tried it with the MLP from this problem (~60 parameters) and it does look as if we're just 20 ms apart. Which makes sense, given the overhead that goes into the computation of `jacfwd` - wrapping the function, handling more diverse types, and so on.

Accordingly, I have removed the custom implementation and replaced it with `jax.jacfwd`.

optimistix/_solver/gradient_methods.py
patrick-kidger: Okay! I guess this isn't really a Jacobian here. Maybe `autodiff`? `autodiff_mode`? Something like that.

johannahaffner:
Summary: 

- After some extra checks, I realised that the compilation-time overhead from using `jax.jacfwd` is additive (20 ms), and apparently independent of problem size. We now use `jax.jacfwd`. 
- the option is now called `autodiff_mode`. Overall, I think that is clearest.
- I added a test to confirm that the forward-branch is entered correctly and that we get the expected result for a trivial case involving `dfx.ForwardMode`. This adds diffrax as a test-time dependency.

tests/test_minimise.py
```diff
@@ -192,3 +197,17 @@ def f(x, _):
...
     assert num_called_so_far == num_called
+
+
+@pytest.mark.parametrize("solver", minimisers)
+@pytest.mark.parametrize(
+    "fn, y0, options, expected", forward_only_fn_init_options_expected
+)
+def test_forward_minimisation(fn, y0, options, expected, solver):
+    if isinstance(solver, optx.OptaxMinimiser):  # No support for forward option
+        return
```
patrick-kidger, line 208: Nit: it'd be better to do
```
@pytest.mark.parametrize("solver", [solver for solver in minimisers if not isinstance(solver, optx.OptaxMinimiser)]
```

Not important enough to block merging, just noting in passing!

patrick-kidger:
Awesome! Merged :D

tests/test_minimise.py
johannahaffner: Noted for the future :)

PR #115, opened 2025-02-14
remove test redundant test case

johannahaffner:
`BFGSBacktracking` is the exact same solver as `optx.BFGS`. The only difference is that the specified `NewtonDescent` does not explicitly use `lx.Cholesky`. But since we tag the Hessian operator as positive semidefinite [here](https://github.com/patrick-kidger/optimistix/blob/0fcce03c3f2533e6df6cff9c9835e0110582cbdb/optimistix/_solver/bfgs.py#L104), lineax will still [pick](https://github.com/patrick-kidger/lineax/blob/fa6c777ef6e5edf2c798a58994e07946a3ce50d7/lineax/_solve.py#L582) Cholesky as a solver.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -148,18 +148,6 @@ class BFGSDogleg(optx.AbstractBFGS):
     verbose: frozenset[str] = frozenset()
 
 
-class BFGSBacktracking(optx.AbstractBFGS):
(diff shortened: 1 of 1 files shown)


patrick-kidger:
LGTM!

Issue #116, opened 2025-02-21
Proximal gradient and L-BFGS


bagibence:
Hi! Thank you for making Optimistix!

We're currently relying on JAXopt in [NeMoS](https://nemos.readthedocs.io/en/latest/), but we're looking to transition to Optimistix (+ Optax).
For this, we would need implementations of proximal gradient descent and L-BFGS. We saw that there are [plans to implement L-BFGS](https://github.com/patrick-kidger/optimistix/issues/3) and some [discussion about constrained optimization](https://github.com/patrick-kidger/optimistix/issues/45).
Are these currently in development by any chance? In case you're interested, we'd be happy to contribute.
We're still getting familiar with the libraries and would appreciate guidance on how best to approach this. Some initial thoughts:
- Proximal gradient: I made a toy implementation for Lasso and Ridge that chains `optax.sgd` and a step imitating Optax's projections which seems to work with `optimistix.minimise`. Do you think this makes sense, or should the proximal operator be handled inside Optimistix instead?
- L-BFGS: In my experiments, wrapping `optax.lbfgs` with `optimistix.OptaxMinimiser` runs without Optax’s line search but fails when it’s enabled.
    - Would it make sense and be feasible to turn `OptaxMinimiser` into a descent object, then combine it with Optimistix’s searches?
    - Maybe modify the wrapper to support Optax’s line search transformations?
    - Or drop Optax and write the required descent step in Optimistix?

@BalzaniEdoardo, @billbrod, @sjvenditto, @gviejo

johannahaffner:
Hi @bagibence, 

yes, constrained optimisation is currently in development! We have a few new descents and searches implemented. 
L-BFGS is also on the list, but hasn't been something I have actively worked on while implementing the constrained searches and descents. 

How about we discuss these plans on a video call?

bagibence:
Thank you for the quick response, that sounds fantastic!

Absolutely, I will send you an email about the video call.

johannahaffner:
Perfect! Have you found my ETH email?

PR #119, opened 2025-03-05
Bounded + Constrained optimisation

johannahaffner:
This is the current state of my development branch, shared here as a basis for discussion. 

I still have some feature branches (second-order corrections in IPOPT, constrained trust regions, OSQP) that I'm going to merge into this once they are a little cleaner.

This also addresses https://github.com/patrick-kidger/optimistix/issues/116, by including a projected gradient descent, similar to optax' implementation of projections.


johannahaffner:
Tests fail because they require the generalised `lx.DiagonalLinearOperator` that takes a PyTree, rather than a vector, as its input (added in https://github.com/patrick-kidger/lineax/pull/125). The latest lineax release predates this addition.

patrick-kidger:
This is awesome. I'll start taking a look through it.

Good point about Lineax, we should do a release. Send a version bump PR and I'll approve?

johannahaffner:
Note: tests pass locally, but I'm experimenting with small breaks to the descent API and pyright is not happy. Looks like we also need to upgrade pyright while we're here.

cjchristopher:
> Yikes, that's a bit scary. That probably means using papers and other implementations as references here.

I don't intend to cause a panic haha. But this is one such issue from jaxopt https://github.com/google/jaxopt/issues/620 l 
I observed similar behaviour with some of the warnings reported in that issue. Where the zoom would end up throwing the point some way outside the bounds, but in jaxopt, from memory, due to how and where the bounds were enforced, this could cause all sorts of bizarre behaviour. Exploding gradients is something I saw a few times. A suggestion on another issue referenced in the one above was to allow it to fail, but this also didn't work.

In any case, I guess this is just a gentle nudge towards rigor and reference - notably, if possible, making sure behaviours roughly matches that of known good implementations elsewhere (numerical quirks notwithstanding)

PR #122, opened 2025-03-12
Support Optax solvers that include a linesearch

### LINKED ISSUE #121: OptaxMinimiser does not work with Optax-Linesearch
Hey, 

I was just trying to use `optimistix.OptaxMinimiser()` with the L-BFGS optimizer from optax. This however does not work, because the included linesearch requires additional inputs to the update-function. The code below results in the error: 

> TypeError: scale_by_backtracking_linesearch.<locals>.update_fn() missing 3 required keyword-only arguments: 'value', 'grad', and 'value_fn'

```
import optimistix
import optax
import jax.numpy as jnp


def test_func(x, args):
    return jnp.sum(jnp.abs(x)**2)

# L-BFGS
solver=optimistix.OptaxMinimiser(optax.lbfgs(learning_rate=1), rtol=1e-6, atol=1e-6)

# Adam + Linesearch
solver=optimistix.OptaxMinimiser(
    optax.chain(optax.adam(learning_rate=1), optax.scale_by_backtracking_linesearch(25)), 
    rtol=1e-6, atol=1e-6)

solution=optimistix.minimise(test_func, solver, y0=jnp.ones(10))
```

I think fixing this should be relatively easy because the other optax optimizers dont care if additional kwargs are provided. E.g. below the `optax.adam()` optimizer runs regardless wether the inputs are provided or not.

```
start_learning_rate = 1e-1
optimizer = optax.adam(start_learning_rate)
params = jnp.ones(10)
opt_state = optimizer.init(params)


for _ in range(1000):
  value, grads = jax.value_and_grad(test_func)(params, args=None)

  updates, opt_state = optimizer.update(grads, opt_state) # standard usage
  #updates, opt_state = optimizer.update(grads, opt_state, params, value=value, grad=grads, value_fn=value) # with linesearch-inputs

  params = optax.apply_updates(params, updates)
```

Anyway since optimistix has its own linesearch options this is nothing crucial. But maybe there are also other optax-features which cause similar issues with `OptaxMinimiser()`?

bagibence:
Needed to pass needed keyword arguments to Optax solvers, and make a new function for the `value_fn` to work.

Currently this works with `optax.scale_by_zoom_linesearch` and `optax.lbfgs` which uses that, but does not work with `optax.scaly_by_backtracking_linesearch`. That needs a fix in Optax that I think I have working, and I will send a PR to them about it.

I extended the list of minimizers in the tests. L-BFGS is tested for least squares as well, explicitly chaining SGD + line search is not.
I also added SGD with backtracking linesearch commented out. After the fix on the Optax side, including this should also pass.

Fixes #121

### DIFF
--- a/optimistix/_solver/optax.py
+++ b/optimistix/_solver/optax.py
@@ -97,7 +97,14 @@ class OptaxMinimiser(AbstractMinimiser[Y, Aux, _OptaxState], strict=True):
                 ("loss" in self.verbose, "Loss", f),
                 ("y" in self.verbose, "y", y),
             )
-        updates, new_opt_state = self.optim.update(grads, state.opt_state, y)
+
+        # fix args and discard aux
+        _fn_for_optax = lambda y: fn(y, args)[0]
+        _fn_for_optax = eqx.filter_closure_convert(_fn_for_optax, y)
+
+        updates, new_opt_state = self.optim.update(
+            grads, state.opt_state, y, value=f, grad=grads, value_fn=_fn_for_optax
+        )
         new_y = eqx.apply_updates(y, updates)
         terminate = cauchy_termination(
             self.rtol,
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -171,7 +171,7 @@ _lsqr_only = (
 
 
 atol = rtol = 1e-8
-minimisers = (
+_general_minimisers = (
     optx.NelderMead(rtol, atol),
     optx.BFGS(rtol, atol, use_inverse=False),
     optx.BFGS(rtol, atol, use_inverse=True),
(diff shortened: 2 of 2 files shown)


bagibence:
The tests pass locally with the latest development version of Optax, but will fail with the current release of 0.2.4 which the workflow also uses.

optimistix/_solver/optax.py
```diff
@@ -97,7 +97,14 @@ def step(
                 ("loss" in self.verbose, "Loss", f),
                 ("y" in self.verbose, "y", y),
             )
-        updates, new_opt_state = self.optim.update(grads, state.opt_state, y)
+
+        # fix args and discard aux
+        _fn_for_optax = lambda y: fn(y, args)[0]
+        _fn_for_optax = eqx.filter_closure_convert(_fn_for_optax, y)
```
patrick-kidger, line 103: I'm surprised the closure-conversion is necessary. Optax itself doesn't return optimizers that are pytrees, is it really expecting that `value_fn` be both a callable and a pytree?

patrick-kidger (commented):
This basically LGTM! Thanks for putting this together.

Give me a ping here once the next Optax release is out and I'd be happy to merge this then.

optimistix/_solver/optax.py
bagibence: You're right, it's not needed. I just removed it.

PR #125, opened 2025-03-17
feat: add DFP quasi newton method

### LINKED ISSUE #123: Help with adding (quasi-newton) Davidon–Fletcher–Powell update method?
Dear optimistix team,

I'm looking into implementing the [`Minuit`](https://www.sciencedirect.com/science/article/pii/0010465575900399?via%3Dihub) optimizer with optimistix. Python bindings for the (newer) C++ implementation can be found in [`iminuit`](https://github.com/scikit-hep/iminuit). 
This optimizer is a standard in high-energy physics research and commonly used by physicists at CERN. Our community is currently in the process of trying to port some of our statistical inference tools (e.g. [`pyhf`](https://github.com/scikit-hep/pyhf)) towards JAX, and thus like to re-implemented Minuit in JAX aswell.

My (best) understanding of `Minuit` is that it consists of 2 parts (that - if specified by a user - are automatically switched between):
1. "MIGRAD": basically the same as [`optimistix.BFGS`](https://github.com/patrick-kidger/optimistix/blob/main/optimistix/_solver/bfgs.py#L320), but the update of the hessian (or its inverse) is done by the [Davidon-Fletcher-Powell formula](https://en.wikipedia.org/wiki/Davidon–Fletcher–Powell_formula) instead.
2. "SIMPLEX": this looks to me exactly like [`optimistix.NelderMead`](https://github.com/patrick-kidger/optimistix/blob/main/optimistix/_solver/nelder_mead.py#L91) - so, perfect!

My wish is to implement MIGRAD in optimistix and for this I had a look at the BFGS code. As far as I understand I probably want to keep the whole code and only switch out [`bfgs_update`](https://github.com/patrick-kidger/optimistix/blob/main/optimistix/_solver/bfgs.py#L79-L106) by a new `dfp_update` function. 
Is this assumption correct? If yes, could you give me some hints what things need to be changed? (I have some troubles understanding the PyTree manipulations in the existing `bfgs_update` code and connect them to the mathematical formula for BFGS.)

In addition, the termination of the minimization is determined in Minuit by the so-called "edm"-measure, which is calculated by: $\text{edm} =\nabla^T H^{-1} \nabla$, where $H^{-1}$ denotes the inverse hessian, see p. 348 in [`Minuit`](https://www.sciencedirect.com/science/article/pii/0010465575900399?via%3Dihub). I suppose I'd have to add this custom measure and the termination logic inside the `step` function aswell?

I'd be happy for any tips and help to make this happen, and contribute this (e.g. the `dfp_update` formula) to optimistix.

Best, Peter

cc @nsmith- and @matthewfeickert

pfackeldey:
Hi @johannahaffner,

as discussed earlier I added the `update_formula` argument in order to switch between `DFP` and `BFGS`. 
As both are of class "quasi-Newton methods" I went ahead and changed the abstract base class name, and renamed the source file accordingly.

I wasn't sure how granular you'd like to have the type signature of the `bfgs_update` and `dfp_update` functions, I went with a simple blank `Callable` for now.

I also added `optx.DFP` to the test suite (but no custom variants as it's currently done for BFGS).

I've also not yet added the `termination criteria` argument - I thought it make sense to separate these into 2 different PRs as it has not really anything to do with the `DFP` algorithm.

Let me know what you think! 

Best, Peter

(closes #123)


optimistix/_solver/quasi_newton.py
```diff
@@ -72,62 +72,97 @@ def leaf_fn(x):
...
+        )
+        return hessian, None
+
+
+_Hessian = TypeVar(
+    "_Hessian", FunctionInfo.EvalGradHessian, FunctionInfo.EvalGradHessianInv
+)
+
+
+def _update_hessian(update_formula, f_eval, grad, prev_grad, y_diff, hessian, hessian_inv) -> _Hessian:
```
johannahaffner, line 147: Can we define `bfgs_update` and `dfp_update` such that they have the same signature (take the same arguments)? They don't have to use all the same arguments, but they should accept them.

Then we don't need this wrapper function here, which will allow us to more cleanly introduce other flavours of quasi-Newton updates in the future, such as for SLSQP. 

I'd like this to be something like

```python
class AbstractBFGS(...):
    ...
    update: AbstractQuasiNewtonUpdate
    ...
    def step(...):
        ...
        def accepted(descent_state):
            ...
            f_eval_info = self.update(...)
            ...
```
where `self.update` takes the same parameters, regardless of the formula used. This could be achieved with an abstract base class, such as 

```
class AbstractQuasiNewtonUpdate(eqx.Module, strict=True):
       use_inverse: bool
       
       @abc.abstractmethod
       def _update(self, ...):
             """Update formula...."""
             
       def __call__(self, ...): 
             return _update(...)
```
This would require some ironing out of the input arguments that allow us to support all the flavours we want, without boxing ourselves in. This probably means switching to passing versions of `FunctionInfo`, which carry most of the attributes, rather than passing everything individually. For example, we could pass `f_info: Union[FunctionInfo.EvalGradHessian, FunctionInfo.EvalGradHessianInv]`, and `f_eval_info: FunctionInfo.EvalGrad`, plus `y, state.y_eval` and that should have us covered for the cases I can think of right now.
It would also define a clean interface others can use to implement their own favourite version. WDYT?

optimistix/_solver/quasi_newton.py
```diff
@@ -170,6 +205,7 @@ class AbstractBFGS(
     use_inverse: AbstractVar[bool]
     descent: AbstractVar[AbstractDescent[Y, _Hessian, Any]]
     search: AbstractVar[AbstractSearch[Y, _Hessian, FunctionInfo.Eval, Any]]
+    update_formula: AbstractVar[Callable]
```
johannahaffner, line 208: How about `hessian_update`? It is not immediately clear what is being updated.

optimistix/_solver/quasi_newton.py
```diff
@@ -235,8 +271,8 @@ def accepted(descent_state):
             else:
                 hessian = state.f_info.hessian
                 hessian_inv = None
-            f_eval_info = _bfgs_update(
-                f_eval, grad, state.f_info.grad, hessian, hessian_inv, y_diff
+            f_eval_info = _update_hessian(
```
johannahaffner, line 274: See above - I think it would be a little neater to call a `self.update` that takes a defined list of arguments and can be swapped without requiring changes to the wrapper function.

optimistix/_solver/quasi_newton.py
```diff
@@ -72,62 +72,97 @@ def leaf_fn(x):
...
-            return hessian, None
 
-    def no_update(hessian, hessian_inv):
-        return hessian, hessian_inv
+def dfp_update(inner, grad_diff, y_diff, hessian, hessian_inv):
+    """https://en.wikipedia.org/wiki/Davidon–Fletcher–Powell_formula"""
+    if hessian is None:
+        # Inverse Hessian
+        assert hessian_inv is not None
+        # DFP update to the operator directly
```
johannahaffner, line 84: This looks good to me! Just `inner` should be moved into this function proper if we switch to a unified signature.

optimistix/_solver/__init__.py
```diff
@@ -37,6 +36,13 @@
     polak_ribiere as polak_ribiere,
 )
 from .optax import OptaxMinimiser as OptaxMinimiser
+from .quasi_newton import (
+    AbstractQuasiNewton as AbstractQuasiNewton,
+    BFGS as BFGS,
+    bfgs_update as bfgs_update,
```
johannahaffner, line 42: Small thing: I'm not sure if the update formulas themselves need to be public given that they already define a solver, but no strong feelings on this.

docs/api/minimise.md
```diff
@@ -32,9 +32,9 @@ In addition to the following, note that the [Optax](https://github.com/deepmind/
 
 ---
 
-??? abstract "`optimistix.AbstractBFGS`"
+??? abstract "`optimistix.AbstractQuasiNewton`"
```
johannahaffner, line 35: I'm a little undecided on renaming this. `AbstractBFGS` is part of the public API, and users might have defined their own custom solvers atop of it, which makes me hesitant to change the name. On the other hand, it is mathematically and conceptually cleaner, and we do have at least one other quasi-Newton update formula planned (SLSQP). I'd appreciate your perspective on this, @patrick-kidger!

johannahaffner (commented):
Nice! Great to get this going. 

The update formulas look good to me, and I think this is a nice way to extend the class to other flavours of quasi Newton updates. To make this process even easier in the future, I suggest we settle on a signature for the update formulas, perhaps encoded through an abstract base class. They should all take the same inputs and return the same outputs, so that users can swap in theirs without having to make changes to the wrapper function. I left some notes about this :) 

I also agree on making this a separate PR - we can deal with the termination conditions later, and this keeps it focused. Thanks! 

Lastly, I'm uncertain about changing the name, since `AbstractBFGS` is part of the public API. Willing to be convinced here!

optimistix/_solver/quasi_newton.py
pfackeldey: That's a great idea, I'll add this! :+1:

optimistix/_solver/quasi_newton.py
pfackeldey: good point! I'll update it :+1:

optimistix/_solver/__init__.py
pfackeldey: I wasn't 100% sure about this at first either, but I think they have to in order to construct custom solvers. Users need to be able to pass these functions to custom solvers for the new arguments - that's why I exposed them.

docs/api/minimise.md
pfackeldey: I'd like to add that I'd definitely would vote for updating the name of the internal state. It would be quite confusing to use `optx.DFP` and work with a `BFGS` state. 
For consistency (and clean interface) I think it would be nice to rename the abstract base class aswell. Maybe we could instead keep `AbstractBFGS` in addition for backwards compatibility (and drop it at some point in a future release)?

johannahaffner: I see your point! And we could either use a type alias or an import alias for backward compatibility. Ideally decorated with a deprecation warning.

optimistix/_solver/__init__.py
johannahaffner: Makes sense!

pfackeldey:
Hi @johannahaffner,

I've updated the code following your suggestions. _This_ implementation was the best I could come up with. 

Could you give it another look? Thanks 🙏

optimistix/_solver/quasi_newton.py
pfackeldey: Done! but...

Unfortunately, I couldn't find a clean way to provide `use_inverse` as an argument to the update formulas as this would e.g. allow passing `use_inverse=True` to the solver, but `use_inverse=False` to the update formula which would be a broken state.

I'm also not 100% sure it the function signatures are as you envision them, e.g. I couldn't use `f_eval_info` (because that's the output of _this_ hessian update).

optimistix/_solver/quasi_newton.py
pfackeldey: done :+1:

optimistix/_solver/quasi_newton.py
johannahaffner: I think if we have `use_inverse` in the update, then we should no longer have it in the solver. It only makes sense to have it in one place, I think. 
Right now we only use it to determine how to compute the update, the descents decide what to do based on the function information they are provided. So I think that makes it a good candidate to live in the update class! Or do you find it more ergonomic to make it an argument that the update takes instead?

Regarding the flavour of function information to pass here - I would go with `FunctionInfo.EvalGrad`, look at how we create it on the fly for the search: https://github.com/patrick-kidger/optimistix/blob/482f9a26d8e3b5f2cc3927ffc8fc9d9b554c24ad/optimistix/_solver/bfgs.py#L224

pfackeldey: > Right now we only use it to determine how to compute the update, the descents decide what to do based on the function information they are provided. So I think that makes it a good candidate to live in the update class! Or do you find it more ergonomic to make it an argument that the update takes instead?

I can do that, but it would break the interface of `BFGS` once more. I think this change makes sense though! Are you ok with that?

pfackeldey: > Regarding the flavour of function information to pass here - I would go with FunctionInfo.EvalGrad, look at how we create it on the fly for the search:

Ok, I'll have a look :+1:

johannahaffner: I think we're already making a break here anyway, so if you agree that it makes sense that let's move that into the update class.

optimistix/_solver/quasi_newton.py
```diff
@@ -72,13 +73,110 @@ def leaf_fn(x):
...
+)
+
+
+class AbstractQuasiNewtonUpdate(eqx.Module, strict=True):
+    """Abstract class for updating the approx. Hessian in a quasi-Newton method."""
+
+    def no_update(self, use_inverse, inner, grad_diff, y_diff, f_info):
+        if use_inverse:
+            return f_info.hessian_inv
+        return f_info.hessian
```
johannahaffner, line 87: Nit: can we use an else here for readability?

optimistix/_solver/quasi_newton.py
```diff
@@ -72,13 +73,110 @@ def leaf_fn(x):
...
+
+    @abc.abstractmethod
+    def update(
```
johannahaffner, line 93: Can we move all of the logic into the update method? (And then call `filter_cond` on inner functions.) We can also do the conversion to lineax operators in there, I think. 

The reason for this is that computing `y_diff`, `grad_diff` as well as the inner product don't necessarily generalise to all quasi-Newton updates, e.g. to those that work with limited memory (L-BFGS), or the ones that are modified to account for constraint information (SLSQP). These should still work with the appropriate flavour of function information. 

I should sit down with pen and paper to confirm that last point (pretty sure though!), I will do that later tonight.

johannahaffner: So long story short: keep `__call__` as generic as possible.

optimistix/_solver/quasi_newton.py
```diff
@@ -228,21 +309,16 @@ def step(
...
-            y_diff = (state.y_eval**ω - y**ω).ω
-            if self.use_inverse:
-                hessian = None
-                hessian_inv = state.f_info.hessian_inv
-            else:
-                hessian = state.f_info.hessian
-                hessian_inv = None
-            f_eval_info = _bfgs_update(
-                f_eval, grad, state.f_info.grad, hessian, hessian_inv, y_diff
+            f_eval_info = self.hessian_update(
```
johannahaffner, line 347: I like how streamlined this now looks!

johannahaffner (commented):
Nice! I think this is going to be a very nice feature. Left some comments here and there :) 

Do we know anything about the properties of DFP vs BFGS? Specifically if one or the other is more suitable for a certain class of problems. It would be nice to add a sentence or two to their respective documentations, such that users get an idea which one might be a better choice for the problem they are trying to solve.

Likewise, I think we could explicate in `AbstractQuasiNewton` that this class is compatible with different flavours of quasi-Newton updates, and that these may be chosen among implementations of `AbstractQuasiNewtonUpdate`.

optimistix/_solver/quasi_newton.py
pfackeldey: I can make `__call__` an abstract method and move all logic of `update` into the concrete `__call__` implementations directly. That's as generic as it can be I think. Is this what you have in mind?

johannahaffner: That works as well, I think. The main thing would be to not put restrictions on future implementations of this update class.

pfackeldey:
I think I addressed all of your comments now @johannahaffner :)

optimistix/_solver/quasi_newton.py
```diff
@@ -335,49 +473,93 @@ class BFGS(AbstractBFGS[Y, Aux, _Hessian], strict=True):
...
+    `verbose=frozenset({"step_size", "loss"})`.
+"""
+
+
+class DFP(AbstractQuasiNewton[Y, Aux, _Hessian], strict=True):
+    """DFP (Davidon--Fletcher--Powell) minimisation algorithm.
+
+    This is a quasi-Newton optimisation algorithm, whose defining feature is the way
+    it progressively builds up a Hessian approximation using multiple steps of gradient
+    information.
```
johannahaffner, line 518: Could we add something like: 

`[optimistix.BFGS][] is generally preferred, since it is more numerically stable on most problems.`

I have not convinced myself of this in the sense that I derived the formula or anything, but it does seem to be the general consensus in the literature. (Interestingly, BFGS as an alternative to DFP seems to have been developed independently, rather than jointly, by Broyden, Fletcher, Goldfarb and Shanno - and they all published their work in 1970.)

johannahaffner (commented):
Thanks! 

I did go over the requirements we would have for L-BFGS and SLSQP, and it seems that we're good when it comes to the information we need to pass to the update class! Everything can be communicated through the appropriate flavours of function information. (If I run into issues related to dual variables for this, then I'll figure those out myself.)

Lastly, I think it would be nice to add the new `AbstractQuasiNewtonUpdate` to the documentation as well, directly in `minimise.md`. We also list different methods to compute the `NonlinearCG` update there, so I think it would fit nicely. (We could do with some indents though.)

optimistix/_solver/quasi_newton.py
pfackeldey: Sure, I'll add this to the docs :+1:

> (Interestingly, BFGS as an alternative to DFP seems to have been developed independently, rather than jointly, by Broyden, Fletcher, Goldfarb and Shanno - and they all published their work in 1970.)

Interesting, I didn't know about this!

docs/api/minimise.md
johannahaffner: Since the attributes are no longer the same, this would not actually work. We could think about a compatibility function, but since the changes are fairly small and not that many people compose custom solvers, its probably fine to add a concise description on how to update.

optimistix/_solver/quasi_newton.py
```diff
@@ -335,49 +473,96 @@ class BFGS(AbstractBFGS[Y, Aux, _Hessian], strict=True):
...
+        norm: Callable[[PyTree], Scalar] = max_norm,
+        verbose: frozenset[str] = frozenset(),
+    ):
+        self.rtol = rtol
+        self.atol = atol
+        self.norm = norm
+        self.descent = NewtonDescent(linear_solver=lx.Cholesky())
+        # TODO(raderj): switch out `BacktrackingArmijo` with a better line search.
+        self.search = BacktrackingArmijo()
+        self.hessian_update = DFPUpdate(use_inverse=True)
```
johannahaffner, line 552: I went over the tests again and just spotted this - can we revert to the old behaviour regarding `use_inverse` here? That is 

```python
def __init__(self, ..., use_inverse: bool = True): 
     ...
     self.hessian_update = BFGSUpdate(use_inverse=use_inverse)
```
Users should definitely be able to choose among these without having to subclass!

johannahaffner (commented):
Whether to use the inverse or not should be an argument to the class constructor of quasi Newton solvers. Otherwise choosing whether to use the inverse or not will require subclassing. 

Can we revert to the previous behaviour here?

optimistix/_solver/quasi_newton.py
johannahaffner: Once we're reverting here, the solver definitions in `helpers.py` can be simplified again.

docs/api/minimise.md
johannahaffner: Actually we can't do this because we're changing attributes. I think a clear description for how to upgrade is the best way to go.

optimistix/_solver/quasi_newton.py
pfackeldey: Hm, I can do that, but there are 2 arguments in favor of how it's now I'd say:
1. Subclassing becomes a bit weird as it's possible to provide `use_inverse` to the solver _and_ to the hessian update. That is quite confusing I'd say and one may even pass two different booleans to them, which results in a broken state of the solver.
2. The inverse is _only_ used in the update, so there's no logical need for it to exist on the solver. Hm, ok, but the same argument goes for `rtol, atol, norm, ...`, so this is not a strong argument.

One idea could be that `use_inverse` is an argument to the `AbstractQuationNewtonUpdate.__call__`, and not to the class instance itself?

Do you have any ideas on how to re-solve (1.)? I can revert it then :)

johannahaffner: Ah, I see that I should have been more clear! I don't want to revert to making it a solver argument, one to have it be an argument of `__init__`, which is passed on to the update when it is initialised.
This would not affect the behaviour of the abstract class, which is what people should subclass.

pfackeldey: Sounds good! I'll do that :)

docs/api/minimise.md
patrick-kidger: @johannahaffner and I just had a quick about this one. I think I agree -- let's (a) rename to `AbstractQuasiNewton`, and (b) not have any `AbstractQuasiNweton.use_inverse` attribute (since it's no longer needed at the solver level).

I don't have strong feelings on whether we also alias `AbstractBFGS = AbstractQuasiNewton`.

patrick-kidger (commented):
I realise I've been pretty silent on this PR so far. It looks like things are shaping up pretty nicely -- let me know if(/when) you think this is done, and I'll go over it!

johannahaffner:
Can we merge this into a dev branch? I would then add L-BFGS on top of this before we do the next release.

Cheers!

pfackeldey:
Hi @johannahaffner and @patrick-kidger,
I’m currently on vacation. The PR is ready from my side though, so feel free to go over it :)

pfackeldey:
Hi @johannahaffner & @patrick-kidger ,
I can't change the target branch to a dev branch (because I have no rights to create new branches I think). 

Also, I updated this PR with the recent changes in main. From my side, this is ready for review.

pfackeldey:
> I've just updated the target branch to `dev`! Thank you very much for this. I'll aim to do a review of this when I get some time on Sunday :)

Awesome, thanks!

optimistix/_solver/quasi_newton.py
```diff
@@ -72,13 +73,156 @@ def leaf_fn(x):
...
+
+    @abc.abstractmethod
+    def __call__(
+        self,
+        f_eval_info: FunctionInfo.EvalGrad,
+        f_info: Union[FunctionInfo.EvalGradHessian, FunctionInfo.EvalGradHessianInv],
+        y: PyTree,
+        y_eval: PyTree,
+    ) -> Union[FunctionInfo.EvalGradHessian, FunctionInfo.EvalGradHessianInv]:
+        ...
```
patrick-kidger, line 94: As this class (and this abstract method) appears in the docs then it'd be nice to have a docstring for this method.

patrick-kidger (commented):
Alright! I've just done a review of this PR, and I really like it.

I've pushed a new commit to this branch with some minor tidy-ups (twiddly things that are easy to do directly whilst reviewing). The only thing worth noting is a bugfix for `DFP(use_inverse=False)`, as `DFP.__init__` had `self.hessian_update = DFPUpdate(use_inverse=True)` with a literal `True`.

I've left one comment below on a missing docstring, and that's it. If you'd be willing to add something there then I'd be very happy to merge this :)

optimistix/_solver/quasi_newton.py
johannahaffner: I could add this when implementing L-BFGS on top of what we have here (which I suggest is done before the next release), alternatively this is my suggestion: 

For the base class: 

```python
    """Abstract base class for quasi-Newton updates to the Hessian approximation. An 
    update consumes information about the current and preceding step, the gradient 
    values at the respective points, and the previous Hessian approximation and returns 
    an updated instance of [`optimistix.FunctionInfo.EvalGradHessian`][] or 
    [`optimistix.FunctionInfo.EvalGradHessianInv`][], depending on whether we're 
    approximating the Hessian or its inverse.
    """
```
For the abstract method:
```python
        """Called whenever we want to update the Hessian approximation. This is usually 
        in the `accepted` branch of the `step` method of an 
        [`optimistix.AbstractQuasiNewton`][] minimiser.

        **Arguments:**
        - `f_eval_info`: The function value and its gradient at the current (just 
            accepted) iterate `y_eval`
        - `f_info`: The function value, gradient and Hessian approximation at the 
            previous accepted iterate `y`
        - `y`: the previous accepted iterate
        - `y_eval`: the current (just accepted) iterate
        """
```

Whilst I'm here I note that we should type this as `y: Y, y_eval: Y`, instead of with `PyTree`.

johannahaffner: And ultranit: I'd prefer the order of arguments to be `y`, `y_eval`, `f_info`, `f_eval_info`, so that the order is the same (previous, current, previous, current) and we're starting with the values we use to explain what is what, but I'm more than happy to put this in myself :D

patrick-kidger: Agreed all the way through. I've just pushed a new update that (a) squashes all of @pfackeldey's commits together, and (b) amends my previous commit with your suggestions.

@johannahaffner can you do a review as another pair of eyes? 😄 If you think this looks good then I think this is now ready to merge!

@pfackeldey likewise please do take a look and double-check I haven't messed anything up 😁

johannahaffner (commented):
A few small comments, otherwise LGTM! 

I also tried building the documentation locally but got various errata, specifically complaints from hippgriffe/jinja on importing `lx.AutoLinearSolver`. @patrick-kidger maybe docs/requirements.txt needs an update?

optimistix/_solver/quasi_newton.py
pfackeldey: Thanks for the thorough review @patrick-kidger & @johannahaffner! I think it looks great now :tada:

I'll quickly push a fix the order of arguments in the doc string as @johannahaffner suggested and then it's good from my side.

Do you want me to do anything else or take it away from here @johannahaffner ?

johannahaffner: I don't have editing rights so I could only do this once I can branch out from dev - if you don't mind then a fix for the docstring/search combinations in the tests would be good! These predate this PR, but in the interest of our future selves I think clarifications there would be useful.

tests/helpers.py
```diff
@@ -112,51 +112,63 @@ def __init__(
...
+    verbose: frozenset[str] = frozenset()
+
+
+class BFGSTrustRegionHessian(optx.AbstractQuasiNewton):
+    """Standard BFGS (uses hessian, not inverse!) + classical trust region update."""
+
+    rtol: float
+    atol: float
+    norm: Callable = optx.max_norm
+    search: optx.AbstractSearch = optx.LinearTrustRegion()
```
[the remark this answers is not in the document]
pfackeldey: I'll update this :+1: do you want me to add the corresponding test cases for DFP aswell @johannahaffner ?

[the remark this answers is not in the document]
johannahaffner: If you're up for it, why not! Never hurts to have comprehensive tests. That said, we wouldn't learn anything from marrying DFP to a linear trust region, since it's behaviour for that is identical to that of BFGS. 

We could combine it with a classical trust region with `use_inverse=False`, paired with `DoglegDescent`, `DampedNewtonDescent`, and with the `IndirectDampedNewtonDescent`, all of which use the step size to solve for the step. (So three custom solvers.)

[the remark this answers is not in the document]
pfackeldey: done!

optimistix/_solver/quasi_newton.py
pfackeldey: Done :+1: It's ready now from my side :)

tests/helpers.py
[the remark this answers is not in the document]
johannahaffner: Great! Everything in shipshape, then 😍

PR #127, opened 2025-04-16
Update workflows: some versions of Github actions sunset today

johannahaffner:
So it looks like today is the day some actions are sunsetting for good: 

https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts

https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

This PR updates the pre-commit action to a more recent one, but seeing as this is also in maintenance-only mode we should probably migrate to something actively maintained: https://github.com/pre-commit/action

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -40,7 +40,7 @@ jobs:
           python -m pip install -r ./tests/requirements.txt
 
       - name: Checks with pre-commit
-        uses: pre-commit/action@v2.0.3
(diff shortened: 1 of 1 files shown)


patrick-kidger:
I think it'd be pretty easy to just run `pre-commit` from the command line.

PR #128, opened 2025-04-16
Proposed structure for integration of benchmarks with pytest-benchmark

johannahaffner:
This PR is a proposal for what integration of benchmarks could look like. 

I suggest: 

- keeping a separate benchmark tests 
- running these with the usual familiar testing machinery
- keeping them out of the normal CI, at least once the benchmark collection is large enough to make inclusion impractical
- including extra information, such as the number of steps, the result of a nonlinear solve, and so on, which is useful for comparison to other implementations
- benchmarking results can be saved to a JSON file, which enables comparison across versions (check if we still get the same accuracy, if we now need more steps, and so on)

I have not included any compile time benchmarking yet, but I think this thing here can happily live next to differently flavoured pytest-codspeed benchmarks, such as discussed in https://github.com/patrick-kidger/equinox/issues/1001

I'm parsing a collection of CUTEST problems for now, and once you get rid of all the FORTRAN stuff and weird formatting these don't take up that much space and can either live here, or perhaps elsewhere if that is more practical.


tests/conftest.py
```diff
@@ -11,3 +11,24 @@
 @pytest.fixture
 def getkey():
     return eqxi.GetKey()
+
+
+# TODO: this can probably be done in a more streamlined fashion!
+# Skip benchmarks on default test runs. Request these explicitly (e.g. when stress-
+# testing new features).
```
patrick-kidger, line 18: It looks like this can be done with a `skipif`: https://stackoverflow.com/a/43938191/22545467

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,43 @@
+# Running benchmarks to test the performance of our solvers, using the pytest
+# benchmarking extension.
+#
+# To run the benchmarks, simply run `pytest benchmarks` from the root directory.
+# If you'd like to save the results, the flag `--benchmark-save=<file_path>` can be used
+# to save a .json file with stats and additional custom metrics. (Useful to compare
+# performance across versions.)
```
patrick-kidger, line 7: Let's put this in `benchmarks/README.md`

benchmarks/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest
+pytest-benchmarks
```
patrick-kidger, line 2: Let's add this line to the `dev` group in `pyproject.toml`. (Just added in #130, #131) and remove the need for a separate requirements file here.

patrick-kidger (commented):
I think the basic structure of this looks good to me! I like being able to integrate this into pytest.

johannahaffner:
`skipif` works, but then the benchmark folder has to be in `tests`, since otherwise we cannot run `pytest benchmarks` separately. In that case it won't look in the `tests` folder, where the extra option is configured. I think it's fine to have the benchmarks folder be a sub-directory of `tests`, so I moved it. 

I also added an abstract base class for benchmark problems, and implemented this for our very favourite Rosenbrock function.

CONTRIBUTING.md
```diff
@@ -38,6 +38,15 @@ pip install -e '.[dev]'
 pytest
 ```
 
+If your changes could affect solver (or compilation) performance, please run the
+benchmark tests with 
+
+```bash
+pytest --extensive
```
patrick-kidger, line 45: Maybe `--benchmark`?

johannahaffner: I think there is to much confusion with the built-in flags then? We already have `--benchmark-save`, which is built-in and specifies where to save the results in benchmark format, and we also use the built-in benchmarking fixture with `@pytest.mark.benchmark`. 

I think that means that our own decorator and flags should be different! Maybe optx_benchmarks?

johannahaffner: Or `@cutest` and `--cutest`? This will work fine until we integrate other benchmarking suites.

johannahaffner:
I'm now going with `@cutest` and benchmarks now have their own configuration. Both might be useful, since we might want to run different benchmarking suites (once we get there!) independently of one another, if they are geared toward different problems and require different solvers.

And we might want to tweak the configuration - to check how well we do in 32-bit, for instance - without affecting the main test suite.

CONTRIBUTING.md
patrick-kidger: You're anticipating a need to separate out solver benchmarks (cutest) vs compile-time benchmarks?

Anyway given that this is now in a separate directory, I'm realising that I think these will be ignored by pytest by default, and only ran if you're in the directory or if you run `pytest benchmarks` from the top-level, so perhaps we don't need a flag here at all?

benchmarks/cutest/problem.py
```diff
@@ -0,0 +1,62 @@
+import abc
+from typing import Any, Union
+
+import equinox as eqx
+from jaxtyping import ArrayLike, PyTree, Scalar
+
+
+_Out = Union[Scalar, PyTree[ArrayLike]]
```
patrick-kidger, line 8: Nit, as we're now Py3.10+ then I'm using `Scalar | PyTree[ArrayLike]` now.

([pyupgrade](https://github.com/asottile/pyupgrade) is a good tool for doing these kinds of conversions automatically.)

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,40 @@
...
+from .helpers import cutest_unconstrained_problems
+
+
+cutest = pytest.mark.skipif("not config.getoption('cutest')")
+
+
+def block_tree_until_ready(x):
+    dynamic, static = eqx.partition(x, eqx.is_inexact_array)
+    dynamic = jtu.tree_map(lambda x: x.block_until_ready(), dynamic)
+    return eqx.combine(dynamic, static)
```
patrick-kidger, line 15: FWIW the partition/combine operations will add a small but measurable amount of overhead, as this happens outside JIT.

I'd be tempted to suggest instead that outputs should be required to always only be JAX arrays, and to then call `jax.block_until_ready`.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,40 @@
...
+@cutest
+@pytest.mark.benchmark
+@pytest.mark.parametrize("problem", cutest_unconstrained_problems)
+@pytest.mark.parametrize("minimiser", unconstrained_minimisers)
+def test_runtime_unconstrained_minimisers(benchmark, minimiser, problem):
+    compiled = eqx.filter_jit(
+        eqx.Partial(
+            optx.minimise, problem.objective, minimiser, problem.y0(), problem.args()
+        )
+    )
```
patrick-kidger, line 31: Likewise for benchmarking I think we should set up things using only `jax.jit`. We've done a pretty good job of optimising `eqx.filter_jit`, but using it is still explicitly an acknowledgement of being will to train a few microseconds of performance in favour of a simpler API.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def wrapped():
+        return block_tree_until_ready(compiled())  # Returns an optx.Solution
```
patrick-kidger, line 34: Given that this consumes no arguments, then at least in principle it may be possible for XLA to constant-propagate at compile time all the way to the solution.

In practice I think it will only do that somewhat, but best to avoid this case I think.

CONTRIBUTING.md
johannahaffner: No, they're not ignored by default, even if they are in a separate directory. 

They would be ignored if the names did not start with `test_`, but then we'd have the reverse problem.

I think runtime and compile time benchmarks can be run with the same problem / solver setup, but the `benchmark` fixture stores all the results and I think things might get muddled there? But happy to try.

benchmarks/cutest/problem.py
johannahaffner: Indeed, I have a visual preference for that too. Should we do that everywhere to be consistent?

benchmarks/test_benchmarks.py
johannahaffner: We could return the value of the objective function at the solution and use that to determine if the problem has been solved to some degree of accuracy.

This is all that is available for 90+ % of CUTEST problems anyway.

More detailed metrics such as the number of iterations could be analysed separately, but we'd lose access to them here.

benchmarks/test_benchmarks.py
johannahaffner: You mean that all but the first run are going to be no-ops? Do I have to make it take array arguments to avoid that?

benchmarks/cutest/problem.py
patrick-kidger: Ah, looks like I'm telling you porkies and that I haven't actually upgraded Optimistix in this way yet (I have started doing so on some of my other libraries.)

I think anything is fine then -- happy to run pyupgrade before or after this PR (but not in this PR, to avoid making it about two things), whicheer is easiest for you.

benchmarks/test_benchmarks.py
patrick-kidger: At least in principle they could be.

In practice I think XLA is a bit milquetoast about actually doing constant propagation to that degree, but I could definitely still see it changing results.

benchmarks/cutest/problem.py
johannahaffner: Haha alright! Let's postpone that for a bit then.

benchmarks/test_benchmarks.py
johannahaffner: Ah you mean only returning the dynamic part of the solution? I understood this as doing something like `return solution.state.f_info.f` or something!

benchmarks/test_benchmarks.py
johannahaffner: Got it!

pfackeldey:
Hi @johannahaffner,
not sure if these thoughts are useful for you here, but:
- I think Github actions workers are not necessarily very homogeneous afaik. So it might be that you get different runtime values if you happen to land on a different worker. Of course, it's always better to have a server that you can use with a reliable small uncertainty on the runtime.
- you might want to restrict JAX to use a single thread only. I ran some benchmarks in the past and found it hard to restrict this (maybe it's better by now?). At least on linux you can use `taskset -c 0 pytest --cutest` to enforce this.

johannahaffner:
They are! Thank you :) 

> Github actions workers are not necessarily very homogeneous 

I was completely unaware of this! But that makes sense. I think depending on the size this benchmark set grows to (fingers crossed), we might not want to integrate it into the CI, but actually run it locally and add some scripts to help analyse the results. So this is another point in favour of maybe not doing this as part of our CI. 

> you might want to restrict JAX to use a single thread only.
And good point! We should take this into account when comparing results to other implementations.

benchmarks/test_benchmarks.py
johannahaffner: So we need at least `solution.RESULTS` which does not support `block_until_ready`. To accommodate this, I wrote a minimal wrapper that calls the compiled function, calls `block_until_ready` on the objective value and returns some extra stuff as well. 

Since the function does not actually do anything else, the overhead should be minimal. I don't know how it compares to a tree-mapped `block_until_ready`, though. If we really want to trim the microseconds then we could use `success = solution.result == optx.RESULTS.successful`, which is an array. 

I suggest we do that when benchmarking for publication, but otherwise it seems overkill - the usage of the benchmarking set is to catch regressions and test performance of new features, and in this case I think it is nice to have a .json file from which we can read informative results, not just minimal timing results.

CONTRIBUTING.md
johannahaffner: I took a quick look at the structure of the `.json` output, and it seems wiser to have two different test functions for compile and runtime. Keeps things neater.

johannahaffner:
I'm introducing a benchmark-time dependency on [sif2jax](https://github.com/johannahaffner/sif2jax) here, which is a current WIP of mine aiming to (have LLMs) port the complete CUTEst collection to clean, human-readable and functionally pure JAX implementations. 

I'll mark this as ready for review again when we have done further verification of the benchmark problems; currently about 247 benchmark problems pass tests against the Fortran implementation.

johannahaffner:
> * you might want to restrict JAX to use a single thread only. I ran some benchmarks in the past and found it hard to restrict this (maybe it's better by now?). At least on linux you can use `taskset -c 0 pytest --cutest` as a "solution".

This does seem to have improved? I've left this in as a commented-out suggestion, but I'm not quite sure what specific behaviour it is meant to prevent if enabled, @pfackeldey.

johannahaffner:
So I tested this a bit more rigorously and it seems like most things are working on our end! 

What we have right now:

- the ability to run benchmarks on  CUTEst problems provided by `sif2jax`, this introduces a benchmark-time dependency on a project of mine
- selectively run on problems up to a certain size
- compare against scipy for specific, defined solvers (other implementations, e.g. in JAXopt, optax, tensor flow can be added following the same pattern)
- analyse benchmark results - all the pytest-benchmark features, e.g. for comparison across runs, saved results with branch, commit info, time stamp, worktree status
- custom analysis: performance profiles of select solvers

I updated the README to reflect this.

Things that require a bit of thought still: 

- how to specify the solvers we're comparing against - solver tolerances do not necessarily have the same meaning in different libraries
- what additional metadata to store and to assess to check if solver comparisons are truly fair
- things like maybe allowing solver-specific maximum numbers of steps
- how/if to put benchmark results under version control so contributors can compare against each other's runs (pytest-benchmark saves these in folders that are specific to the platform (Darwin, Linux, ...), the Python implementation (CPython, ...), the Python version and the precision used, so no risk of mixing these up). Currently the benchmarking results are git ignored.
- I think it would be nice to ask Contributors that add a new solver if they could specify alternative implementations to benchmark against, since they are probably best positioned to judge the finer details of this -> this can go into CONTRIBUTING.md. Or does that make things sound too scary?

I'll mark this as ready to review for now :)

CONTRIBUTING.md
patrick-kidger: Okay! So in that case maybe `--compile-benchmarks` and `--runtime-benchmarks`?

Or compose with existing pytest functionality, `pytest --benchmark -k compile` vs `pytest --benchmark -k runtime`?

Mostly I'd just like to have the flag make it clear that this isn't about running slow tests or something.

CONTRIBUTING.md
```diff
@@ -38,6 +38,15 @@ pip install -e '.[dev]'
 pytest
 ```
 
+If your changes could affect solver (or compilation) performance, please run the
+benchmark tests with
```
patrick-kidger: +mention that they could run benchmarks before/after their change to test it.

Honestly I am starting to warm to the idea of using a service like codspeed etc to log benchmark results over time.

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
+# Running benchmarks 
+
+We're using [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/) to assess performance. Benchmarks can be run with 
+
+```
+pytest benchmarks/ --cutest
```
patrick-kidger: Ah, looks like this needs syncing with `CONTRIBUTING.md`.

What happens if you only run `pytest benchmarks`?

What stops the benchmarks from being ran on just `pytest`?

How do you want to handle forward compatibility with any of our own benchmarks in the future? I'm anticipating that the `--cutest` flag would then no longer be the right name as we probably do want people to run our own benchmarks at the same time, and not make the distinction.

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+pytest benchmarks/ --cutest --benchmark-save=<file_path>
+```
+
+benchmark results can be saved in a .json file. Additional custom metrics (e.g. the number of iterations or the quality of the solution) can be configured in the benchmark test functions. 
+
+This will just run benchmarks on our own solvers - you can use saved results to compare performance to previous versions or different commits. Comparing against the last saved run is enabled with `pytest --benchmark-compare`, but specific iDs of previous runs may also be specified. (Consult the [documentation](https://pytest-benchmark.readthedocs.io/en/latest/) for more options.)
+
+For reproducibility, make sure that all your changes have been committed and your working tree is clean before running the benchmarks. `pytest-benchmark` will otherwise mark your benchmarking results as `dirty`.
```
patrick-kidger: Does this matter? It's all local anyway.

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+benchmark results can be saved in a .json file. Additional custom metrics (e.g. the number of iterations or the quality of the solution) can be configured in the benchmark test functions. 
+
+This will just run benchmarks on our own solvers - you can use saved results to compare performance to previous versions or different commits. Comparing against the last saved run is enabled with `pytest --benchmark-compare`, but specific iDs of previous runs may also be specified. (Consult the [documentation](https://pytest-benchmark.readthedocs.io/en/latest/) for more options.)
+
+For reproducibility, make sure that all your changes have been committed and your working tree is clean before running the benchmarks. `pytest-benchmark` will otherwise mark your benchmarking results as `dirty`. 
+Saved results include the commit, branch, version, and an exact timestamp by default.
+
+Note that benchmarks are run with `throw=False` enabled, since otherwise no result is written in the json file, but we do want to know if we failed to solve a problem.
```
patrick-kidger: If you wanted we could/should set `EQX_ON_ERROR=nan` here too. This is essentially equivalent to the above and is also a performance improvement.

(This has to happen before Equinox is imported.)

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Benchmark against defined SciPy solvers with `--scipy` (you might want to restrict the dimensions here, SciPy is quite slow).
+
+## Visualising results with performance profiles
+
+Performance profiles for any list of solvers (run on problem sets with overlap) can be generated using profile.py.
+The usage of this is 
+
+```bash
+python3 benchmarks/profile.py solver1 solver2 platform python_version precision iD
```
patrick-kidger: `python3` -> `python`

benchmarks/conftest.py
```diff
@@ -0,0 +1,43 @@
+import jax
+
+
+jax.config.update("jax_enable_x64", True)
+jax.config.update("jax_numpy_rank_promotion", "raise")
+jax.config.update("jax_numpy_dtype_promotion", "standard")
+# Remark Peter: limit JAX to single-thread with taskset -c 0 pytest --cutest
```
patrick-kidger: Note that `taskset` is Linux-only, I believe.

benchmarks/profile.py
```diff
@@ -0,0 +1,228 @@
...
+import os
+import pathlib
+import re
+import sys
+import warnings
+
+import jax.numpy as jnp
+import matplotlib.pyplot as plt  # pyright: ignore[reportMissingImports]
+import numpy as np  # pyright: ignore[reportMissingImports]
+import pandas as pd  # pyright: ignore[reportMissingImports]
```
patrick-kidger: Is pandas that useful here? (It seems fairly minimal below.)

To me this is in the same bucket as libraries like huggingface and lightning - reasonable for hacking around but not a thing I'd like to check in. (I realise this is a somewhat unusual position.)

benchmarks/profile.py
```diff
@@ -0,0 +1,228 @@
...
+            "Please ensure it exists in the current or parent directory."
+        )
+
+    print(f"Using benchmarks from: {benchmarks_dir}")
+    return benchmarks_dir
+
+
+def find_benchmark_run(benchmarks_dir: pathlib.Path, args: list) -> dict:
+    """Find the benchmark run data based on the provided arguments."""
+    platform, python_version, precision, iD = args[-4:]
```
patrick-kidger: Maybe we should use https://github.com/google/python-fire to make the argument-parsing nice?

I think it's okay to introduce the extra dependency as we can put it in the dev dependency list.

pyproject.toml
```diff
@@ -46,7 +46,9 @@ dev = [
   "optax>=0.2.4",
   "pyright>=1.1.399",
   "pytest>=8.3.5",
-  "ruff>=0.11.6"
+  "pytest-benchmark>=5.1.0",
+  "ruff>=0.11.6",
+  "sif2jax>=0.0.3"
```
patrick-kidger: I think this needs extending with pandas (if we keep it) and matplotlib.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,132 @@
...
+            optx.minimise,
+            problem.objective,
+            minimiser,
+            args=problem.args(),
+            max_steps=max_steps,
+            throw=False,
+        )
+    )
+
+    _ = solve(problem.y0())  # Warm up
```
patrick-kidger: I don't think we need the `_ =`

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def wrapped(y0):
+        solution = solve(y0)
+        objective_value = solution.state.f_info.f.block_until_ready()
```
patrick-kidger: Maybe just `jax.block_until_ready(solution)`?

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,132 @@
...
+        num_steps = solution.stats["num_steps"]
+        return objective_value, solution.result, num_steps
+
+    # Benchmark the runtime of the compiled function
+    values = benchmark.pedantic(wrapped, args=(problem.y0(),), rounds=5, iterations=1)
```
patrick-kidger: How does the benchmarking work here btw? As compared to e.g. `min(timeit.repeat(..., number=..., repeat=...))` or the ipython `%%timeit`? (E.g. the former uses minimisation to handle one-sided noise, whilst the latter automatically figures out how many runs to perform.)

patrick-kidger (commented):
I really like the look of this.

CONTRIBUTING.md
johannahaffner: We now use [these](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options) built in flags and skip by default. Regular -k flags also work.
I moved the benchmarking script into the regular tests folder again, since it is now quite small. 

This makes the setup less complex, in particular we are no longer dealing with two different conftest.py files, I had noticed that running with `pytest benchmarks/` was actually needed to pull in the custom flags (e.g. `--max-dimension`). Now there is a single flag to run the benchmarks, and custom ones for subsets.

benchmarks/README.md
johannahaffner: I switched to the built-in flags, these already provide most of what I wanted to go for here. 
Default for benchmark runs is now false, this is set in pyproject.toml and can be overridden with this flag here.

We can always re-introduce a CUTEst flag with a fixture that selects based on a problem attribute if we happen to need it for specific solver comparisons.

I'll add the rest of the information to CONTRIBUTING.md.

benchmarks/README.md
johannahaffner: I think the part about having a clean working tree is relevant if we want to log and compare benchmark results, yes.

benchmarks/conftest.py
johannahaffner: I'm not sure if it what this is meant to prevent/accomplish, if we added it.

benchmarks/profile.py
johannahaffner: Oh I share this view of pandas! It has like one million dependencies, and every requirements file stops being human-readable as soon as this is in. 

It is probably possible to factor out a utility that mostly consists of tree-mapped nanmin though.

benchmarks/test_benchmarks.py
johannahaffner: This would add the overhead of a `tree_map` and for that reason we had decided not to do that. Background [here](https://github.com/patrick-kidger/optimistix/pull/128#discussion_r2072340890) :)

benchmarks/test_benchmarks.py
johannahaffner: So rounds and iterations correspond to repeats and number. I switched to @jpbrodrick89's suggestion of not using the pedantic option.

When saving the result, more than just the minimum gets saved - a bunch of statistics get written to an output `.json`, and these can be selected among during analysis. There is also an option to save full results if desired.

jpbrodrick89:
Hi guys, I will chime in a tiny bit more detail this afternoon as I've been trying a lot of this on one of my repos. Unfortunately CodSpeeds usual instrumentation provides no value for Jax functions as everything counts as a system calls which is not trackable by valgrind. Therefore to use their tooling you have to use their wall runner where costs can mount up pretty quickly. I had 360 micro benchmarks and used up the free 2 hours in just four commits. Also, even on their bare metal wall runner there is still some noise probably due the time of actually retrieving the function from the jax cache rather than running the actual floap. Trying to work out if I can make this more useable next week. At least they've updated their docs in the last couple weeks quite substantially.
@johannahaffner happy to have a call if helpful.

benchmarks/test_benchmarks.py
jpbrodrick89: Typically 5 rounds is very small compared to defaults and will be very susceptible to noise but depends how expensive the calculation is. I typically trust the standard non pedantic option.

benchmarks/test_benchmarks.py
jpbrodrick89: Agreed

benchmarks/test_benchmarks.py
jpbrodrick89: Any reason not to be more explicit with what you're doing and just do `.lower(...).compile()`?

benchmarks/README.md
jpbrodrick89: Note that this is essential for vjp as throw=False does nothing there I believe

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,40 @@
...
+#           python-version: ${{ matrix.python-version }}
+
+#       - name: Install dependencies
+#         run: |
+#           python -m pip install --upgrade pip
+#           python -m pip install '.[dev]' 'jax[cpu]'
+
+#       - name: Run benchmarks
+#         uses: CodSpeedHQ/action@v3
+#         with:
```
jpbrodrick89: This is where you can add the env variables
```suggestion
#         with:
#           env:
#             EQX_ON_ERROR: "nan"
```

benchmarks/README.md
johannahaffner: Argh I see that I might have boxed myself a little into a corner here - I could simplify the workflow and put the benchmarks back in the tests folder. This way we only get one `conftest` and one setup to deal with, and there are no stupid foot guns such as custom flags only being available if we're also specifying the folder name: e.g. 

`pytest benchmarks/ --benchmark-only--max-dimension=10` 

would work, but 

`pytest --benchmark-only --max-dimension=10` 

would not, which I had not previously realised.

But this also means that the environment variable applies to the whole test session, right? And while we want it for benchmarking, we don't want it for testing. 

Can I force a fresh environment for a specific testing module? Alternatively it might make sense to modify `--benchmark-only` so that if present, we do set the environment variable and not otherwise.

jpbrodrick89: Yes I would use a new environment for benchmarks (I find this hepful in my workflow as tests run faster than benchmarks and I can fail fast if something is wrong). Then no need to fiddle around with environment variables in `tests/` or `benchmarks/` and they are just set in the workflow actions (or by user). This is also helpful as it allows the user to run with different setting to compare impact.

johannahaffner: In Github workflows? That makes sense! Locally we default to no benchmark runs, and I just discovered that I can monkeypatch the environment variables for specific tests only - would that work, @patrick-kidger?

pfackeldey:
> I opted against using taskset, because I don't really understand what it would/should do

Fyi, [`taskset`](https://www.man7.org/linux/man-pages/man1/taskset.1.html) lets you restrict the number of cores seen by a process. E.g. `taskset -c 1 python jax_code.py` would _only_ use 1 core and not however many are available on the machine. IIRC jax is rather aggressive about this and uses whatever is available.

I assume this only makes sense if you are interested in having a history of benchmarks over multiple CI runs, _and_ given the premise that Github Actions runner may have different number of cores. (I'm not sure how homogeneous GitHub actions are?).

johannahaffner:
> I assume this only makes sense if you are interested in having a history of benchmarks over multiple CI runs, _and_ given the premise that Github Actions runner may have different number of cores. (I'm not sure how homogeneous GitHub actions are?).
> 
> (This only works on linux, and if the cores differ in clock frequency between different workers, that also doesn't help of course)

Thank you for the explanation! Does this mean that you can get around some of the variability in CI runs? Apparently how fast things are in there depends on a lot of factors. That is actually what codspeed advertises - they claim that they get the noise down from around 30 % to 1 %; and they do that through some clever isolation of the bit to be benchmarked, which is then run just once. But this does, apparently, not really mesh well with JAX. 

FWIW I've come to the conclusion that the pragmatic way forward is to just use the benchmarks locally during development.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,151 @@
...
+            problem.objective,
+            minimiser,
+            args=problem.args(),
+            max_steps=max_steps,
+            # We save on failure and filter during analysis to be able to compute the
+            # fraction of solved problems for performance profiles.
+            throw=False,
+        )
+    )
+
```
jpbrodrick89, line 98:
```suggestion
y0 = jax.tree.map(lambda x: jax.device_put(x) if eqx.is_array(x) else x, problem.y0())
```
This makes sure that device transfer times are not includedi n benchmarks which can be significant for GPU's.

johannahaffner: Thanks, accepted!

johannahaffner:
I think this is ready! Thanks to all of you for your help, @patrick-kidger @jpbrodrick89 and @pfackeldey. 

Small updates made since last synopsis:

- accepted @jpbrodrick89's suggestion to use device put
- added a comment to use task set in `run_tests.yml` - for future reference, once this becomes part of the CI
- bumped the minimum version of `sif2jax` to one that is compatible with strict dtype and rank promotion rules
- moved the benchmarking script back to the `benchmarks` folder, with its own `conftest` - this is to make sure that we have a separate benchmarking environment, which now uses `EQX_ON_ERROR = "nan"` and does not monkeypatch anything.
- updated CONTRIBUTING.md to match.

jpbrodrick89:
Exciting time @johannahaffner! 🚀 It looks like you set eqx on error as a global python variable rather than an os environment variable, see here for different options for setting env variables in pytest https://stackoverflow.com/questions/36141024/how-to-pass-environment-variables-to-pytest

The way I do is just to set on command line with `EQX_ON_ERROR=nan pytest....`

johannahaffner:
> Awesome! Not at the desk right now to to check, but i think you should be able to get rid of the noqa e402s now as os is meant to be ignored (https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/) I'm not sure about the i001would be nice if this had a similar exception but it's not documented.

Indeed! And `I001` does not make exceptions.

johannahaffner:
I tried some benchmarking on our brand new L-BFGS and the current setup, using `jax.jit` exclusively, complains about the jaxpr in its state. I'll see if I can figure out a way around this that also works for compilation benchmarking.

johannahaffner:
@patrick-kidger what do you think about merging this as-is, with the `jax.jit`/ jaxpr in solver state blocker merely documented, but not resolved?

We do have support for runtime benchmarking, we just cannot benchmark the (AOT) compilation time for a subset of solvers. This PR is now a little old - we did think about it some three months ago, which is also when it was mostly done. To be honest, most things were no longer very fresh in my mind, and I had to re-run some benchmarks and generate some figures to convince myself that things are as they should be. 

Merging this with support for runtime benchmarking only seems to be a pragmatic way forward here.

CONTRIBUTING.md
```diff
@@ -38,6 +38,13 @@ pip install -e '.[dev]'
 pytest
 ```
 
+If your changes could affect solver (or compilation) performance, please run the benchmark tests with 
+
+```bash
+pytest benchmarks/ --benchmark-only
```
patrick-kidger, line 44: Is the `benchmarks/` necessary? IIUC then:
- a root-level `pytest` will collect both `tests/` and `benchmarks/`;
- the default `--benchmark-skip` will ignore everything in the latter;
- setting `--benchmark-only` will ignore the skip *and* ignore the `tests/`.

CONTRIBUTING.md
```diff
@@ -38,6 +38,13 @@ pip install -e '.[dev]'
...
+```
+
+You can run benchmarks on `main` / `dev` or on your feature branch for a before-and-after comparison, and also save more extensive results for analysis. For more on this, skip to the "Benchmarking" section below.
```
patrick-kidger: Recognising that some (most) contributors are drive-by and don't engage with the main/dev distinction (and that that's okay), a suggested simplification:
```diff
- You can run benchmarks on `main` / `dev` or on your feature branch for a before-and-after comparison, and also save more extensive results for analysis. For more on this, skip to the "Benchmarking" section below. 
+ You can run benchmarks before or after your change, and also save more extensive results for analysis. For more on this, skip to the "Benchmarking" section below. 
```

CONTRIBUTING.md
```diff
@@ -58,3 +65,40 @@ mkdocs serve
...
+and compare against previous runs with `pytest --benchmark-compare`, which will automatically pull in the last saved commit, but also takes run iDs as arguments. See the `pytest-benchmark` [documentation](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options) for more command line options. 
+The `benchmark-autosave` option will specify the commit iD, instead of a user-defined name.
+Make sure that you are running benchmarks with a clean working tree, so you can trace how changes affect performance!
+
+For convenience, we support some custom flags: 
+
+- `--min-dimension=<int>, --max-dimension=<int>` benchmarks can be run on a subset of problems based on problem size.
+- `--scipy` benchmarks of our solvers are run against the corresponding Python implementation. You might want to limit problem dimension here, they can be quite slow.
+
+pytest' `-k` flags also work in this setting to enable selective execution of benchmarking functions.
```
patrick-kidger:
```diff
- pytest'
+ pytest's
```

CONTRIBUTING.md
```diff
@@ -58,3 +65,40 @@ mkdocs serve
...
+
+**Analysing benchmark results**
+
+You can find a script to analyse benchmark results in `benchmarks/profile.py`. Run it with
+
+```bash
+python benchmarks/profile.py platform python_version precision iD kind *solver_names
```
patrick-kidger: I think platform, python version etc should be automatically detectable?

That aside when writing CLI information, to distinguish between e.g. a literal `python` and a placeholder `platform`, it is common to encapsulate the latter in `<>`, e.g. `python benchmarks/profile.py <platform> ...`

benchmarks/profile.py
```diff
@@ -0,0 +1,306 @@
...
+                # TODO(jhaffner): currently assumes that problems are identifiable by
+                # their names and ignores variably dimensioned problems, where several
+                # problems would have the same name, but different attributes. We do not
+                # yet support this in sif2jax, but support is planned for the short term
+                solver_data[solver_name][problem_name] = {
+                    "min_runtime": min_runtime,
+                    "successful": is_successful,
+                }
+
+    return OrderedDict(solver_data)
```
patrick-kidger, line 154: Note that regular `dict`s are ordered as of Python 3.7 (and CPython 3.6) so this isn't needed.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,216 @@
...
+            solver,
+            args=problem.args(),
+            max_steps=max_steps,
+            # We save on failure and filter during analysis to be able to compute the
+            # fraction of solved problems for performance profiles.
+            throw=False,
+        )
+    )
+    y0 = jax.tree.map(
+        lambda x: jax.device_put(x) if eqx.is_array(x) else x, problem.y0()
```
patrick-kidger, line 95: `lambda x: jax.device_put(x)` can be replaced with just `jax.device_put`.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,216 @@
...
+    # Benchmark the runtime of the compiled function
+    wrapped(problem.y0())  # Warm up
+    benchmark(wrapped, problem.y0())
+
+    # Save information and results
+    benchmark.extra_info["problem name"] = problem.__class__.__name__
+    benchmark.extra_info["solver name"] = name
+
+
+@scipy
```
patrick-kidger: Nit: this decorator is only used once, so I'd suggest just inlining the definition.

docs/_static/quadratic_bowl.py
```diff
@@ -14,19 +14,19 @@ def quadratic_bowl(x, y):
...
-ax.set_zlabel("f(x, y)\n\n\n\n\n\n\n\n", labelpad=-30)
-ax.zaxis.set_rotate_label(False)
-ax.xaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))
-ax.yaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))
-ax.zaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))
+ax.set_zlabel("f(x, y)\n\n\n\n\n\n\n\n", labelpad=-30)  # pyright: ignore
+ax.zaxis.set_rotate_label(False)  # pyright: ignore
+ax.xaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))  # pyright: ignore
+ax.yaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))  # pyright: ignore
+ax.zaxis.set_pane_color((1.0, 1.0, 1.0, 1.0))  # pyright: ignore
```
patrick-kidger: Nit: Can these be `pyright: ignore[somethingSpecific]`?

(I didn't used to do this, but I think it's a best practice.)

patrick-kidger (commented):
Okay nits only! This basically LGTM.

CONTRIBUTING.md
johannahaffner: Almost - 

`--benchmark-only` without the `benchmarks/` would collect everything in `tests/` and then _skip_ it (and not skip the benchmarks). 

`benchmarks/` is necessary for the definition of custom options in `benchmarks/conftest.py`, such as filtering by the minimum and maximum number of dimensions, or running `scipy` tests. If `benchmarks/` is not used, then passing one of our options will result in an error, since they are not defined in the `tests` folder.

CONTRIBUTING.md
johannahaffner: The Python version on which the tests are run is detected automatically and results are saved with the appropriate path. 

Detecting the version from which this script is ran to generate the plots would mean that we assume that these are always the same, which is probably true most of the time but not something I would like to bake in.

benchmarks/profile.py
johannahaffner: I only got rid of a bug after switching to this, on Python 3.13.

Results were switched, and getting a correct comparison of our LBFGS vs. our BFGS and `optax.lbfgs` was only possible after switching to `OrderedDict`.  Not sure why - perhaps because of the `set` operations?

benchmarks/test_benchmarks.py
johannahaffner: Right, but what about the filtering using `eqx.is_array`?

docs/_static/quadratic_bowl.py
johannahaffner: I did it for the ones for which this did work - some of them seem to raise more than one error and after some tinkering I kept `pyright: ignore` for some of these to pass `pre-commit`.

benchmarks/test_benchmarks.py
patrick-kidger: Oh sorry, I'm misreading the lambda. Yup ignore me.

benchmarks/profile.py
patrick-kidger: Hmm, that seems very strange. I would have thought that this call basically does nothing.

Either way FWIW, to me this is suggestive of a code smell. I'd generally argue that it's best to treat dicts as unordered, as depending on their ordering is often a subtle point missed during refactors.

Staring at your code I don't seem an obvious issue though. The `set` operations would result in a random ordering within `get_relative_performance`, but I don't think that matters...

CONTRIBUTING.md
patrick-kidger: Is it worth avoiding the problem by having the benchmarks directly output the results of this script? IIUC then this should be pretty cheap.

(I'm basically just putting my usual UI hat on here – trying to minimise the number of arguments / simplicity for a user, as long as the computational cost isn't too high.)

CONTRIBUTING.md
patrick-kidger: Ah, got it.
Hmm in that case possibly the opposite strategy would work: just `pytest benchmarks` could work, with `benchmarks/conftest.py` adding `--benchmark-only` automatically?

CONTRIBUTING.md
johannahaffner: See reply in https://github.com/patrick-kidger/optimistix/pull/200.

CONTRIBUTING.md
johannahaffner: See comment in https://github.com/patrick-kidger/optimistix/pull/200.

PR #136, opened 2025-05-12
Fixed Gauss-Newton including a spurious `lx.linear_solve(..., throw=True)`

patrick-kidger:
And moreover when checking that the new `throw=False` would have any of its failing cases picked up, I noticed that the downstream `AbstractGaussNewton.terminate` method does not actually check the result when deciding whether to terminate. Rather than adjusting just the one solver, I decided to simply lift this condition into the top-level `_iterate.py`.

As such we might consider whether we really need `.terminate` to return both `termintae: bool` and `result: RESULTS`. To consider what each combination means:

- `terminate=True`, `result=RESULTS.successful`: done, and all went well.
- `terminate=True`, `result=RESULTS.foo`: something went wrong; halt.
- `terminate=False`, `result=RESULTS.successfull`: keep making more iterations.
- `terminate=False`, `result=RESULTS.foo`: <invalid>

So we really do need two return arguments: it's not enough to just stop if `result != RESULTS.successful`, because otherwise we have no way to communicate success-and-terminate.

Now we might wonder whether this change produces a change in behaviour. I think technically yes: a solver may have set `terminate=False` and `result=RESULTS.foo`, and then later decided to switch back to being successful. In practice I don't think we ever did that, and it seems pretty sketchy anyway.

### DIFF
--- a/optimistix/_iterate.py
+++ b/optimistix/_iterate.py
@@ -223,8 +223,8 @@ def _iterate(inputs):
     def cond_fun(carry):
         y, _, dynamic_state, _ = carry
         state = eqx.combine(static_state, dynamic_state)
-        terminate, _ = solver.terminate(fn, y, args, options, state, tags)
-        return jnp.invert(terminate)
+        terminate, result = solver.terminate(fn, y, args, options, state, tags)
+        return jnp.invert(terminate) | (result != RESULTS.successful)
 
     def body_fun(carry):
         y, num_steps, dynamic_state, _ = carry
--- a/optimistix/_solver/gauss_newton.py
+++ b/optimistix/_solver/gauss_newton.py
@@ -73,7 +73,7 @@ def newton_step(
                 "Cannot use a Newton descent with a solver that only evaluates the "
                 "gradient, or only the function itself."
             )
-        out = lx.linear_solve(operator, vector, linear_solver)
+        out = lx.linear_solve(operator, vector, linear_solver, throw=False)
         newton = out.value
         result = RESULTS.promote(out.result)
     return newton, result


optimistix/_iterate.py
```diff
@@ -223,8 +223,8 @@ def _iterate(inputs):
     def cond_fun(carry):
         y, _, dynamic_state, _ = carry
         state = eqx.combine(static_state, dynamic_state)
-        terminate, _ = solver.terminate(fn, y, args, options, state, tags)
-        return jnp.invert(terminate)
+        terminate, result = solver.terminate(fn, y, args, options, state, tags)
+        return jnp.invert(terminate) | (result != RESULTS.successful)
```
johannahaffner, line 227: This means that an unsuccessful step is unrecoverable, right?

So we assume that a direction for which the linear solve result is not successful is never going to yield a direction along which an acceptable step may be found. Is that true in general? I'm also thinking about the lineax results we get from the iterative solvers, which may indicate stagnation or that the maximum number of steps has been reached, IIRC.

johannahaffner (commented):
A counterexample: `terminate = False` and  `RESULTS.max_steps_reached`, which may be returned by iterative solvers in lineax. 
In that case we would want to continue.

But in other cases - nonfinite input, for instance - we do want to terminate. And I also do think that it is attractive to do that as early as we can, without potentially adding a few hundred rejected steps that never compute a new direction and hence cannot recover. (`damped_newton_step` is called in `descent.step`, and so is called in rejected steps too, but that one seems particularly unlikely to fail in such a way, since only the step size changes which is needed to compute $\lambda \mathbb{I}$.)
So an alternative would be to: do more postprocessing of lineax results in the descents or solvers, or to have a number of allowable results and compare against that.

patrick-kidger:
> A counterexample: terminate = False and RESULTS.max_steps_reached, which may be returned by iterative solvers in lineax. In that case we would want to continue.

Why is that? This sounds like we should halt with a failure, as we're currently in a broken state. This sounds like exactly the kind of issue that this check would catch.

johannahaffner:
> > A counterexample: terminate = False and RESULTS.max_steps_reached, which may be returned by iterative solvers in lineax. In that case we would want to continue.
> 
> Why is that? This sounds like we should halt with a failure, as we're currently in a broken state. This sounds like exactly the kind of issue that this check would catch.

Hm not sure? I never use iterative linear solvers, but we do in principle support their usage. Maybe someone wants to be greedy and only use a small number of steps in the iterative linear solve?

That aside, I do use results to request feasibility restoration in the constrained solvers and was toying with the idea of passing that to the next step instead of resolving it immediately within `solver.step`. But we could also agree that whenever we have `result != RESULTS.successful` and the solver can do something about it, whatever is done about it happens within the `step`.

johannahaffner:
Alternative the solvers could decide in `solver.terminate` if `RESULTS.successful` is required, and by default return `state.terminate | result != RESULTS.successful`.

PR #135, opened 2025-05-06
LBFGS Hess approx

BalzaniEdoardo:
## Draft PR: L-BFGS Implementation for `quasi_newton.py`

Hello, and first of all, thank you for the great package!

In this draft PR, I’m working on implementing the **L-BFGS algorithm** within the `quasi_newton.py` module, targeting part of #116 . Before consolidating the code with tests and full integration, I’d appreciate guidance on design decisions.

---

##  Implementation Overview

- The descent direction is computed via `_lim_mem_hess_inv_operator_fn`, which implements the two-loop recursion using the history of parameter and gradient residuals.
- `_lim_mem_hess_inv_operator` acts as an operator factory: it partially evaluates `_lim_mem_hess_inv_operator_fn` with the current residuals and returns a `lineax.FunctionLinearOperator`.
- Currently, the operator is **materialized** before returning, which likely defeats the purpose of using an implicit representation.
- The buffers with the residuals are stored in a pytree of arrays with the same structure as the parameters, but with an additional dimension (of length "buffer size").
- Residuals are stored in a dictionary within `_QuasiNewtonState`.
- The Hessian update returns an additional dictionary carrying the updated residual state.

---

## Questions
1. **Materialization vs. Tree Equality**

    To satisfy this assertion in `_iterate.py`:
    ```python
    assert eqx.tree_equal(static_state, new_static_state) is True
    ```
    I had to materialize the FunctionLinearOperator. Is there a way to avoid this and retain the implicit operator while still passing this check?

2.  The **memory buffer size** is currently **fixed** at 10 iterations. What’s the best way to expose this as a user-settable parameter without altering the broader solver API in `quasi_newton.py`?

3. I did not systematically tested it yet, but for low-dimensional parameters, JIT-compiling the function `_lim_mem_hess_inv_operator_fn` and calling it directly seems significantly faster than the construction of a `FunctionLinearOperator`. Below is line_profiler output for a minimise call:
 
 ```
Timer unit: 1e-06 s

Total time: 0.052245 s
/Users/ebalzani/Code/optimistix/optimistix/_solver/quasi_newton.py
Function: _lim_mem_hess_inv_operator at line 120

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   120                                           @line_profiler.profile
   121                                           def _lim_mem_hess_inv_operator(
   122                                                   residual_par: PyTree[Array],
   123                                                   residual_grad: PyTree[Array],
   124                                                   rho: Array,
[...]

PR #143, opened 2025-06-13
Add zoom linesearch

bagibence:
As we talked about, I implemented zoom linesearch that satisfies the strong Wolfe conditions, which is particularly useful for quasi-Newton methods.

After some iterations, the current version ended up similar to the ones in [Optax](https://github.com/google-deepmind/optax/blob/main/optax/_src/linesearch.py) and [JAXopt](https://github.com/google/jaxopt/blob/main/jaxopt/_src/zoom_linesearch.py), but compatible with Optimistix's flattened loop. It works but definitely has some rough edges, so please let me know what changes you'd like me to make.

The main complication is that to check the curvature condition, we need the slope at `y_eval`, and for that we need the gradient at `y_eval`. I couldn't do that with the current interface of `AbstractSearch.step`, so I modified it to accept `lin_fn` and `options`, which are then used to calculate the gradient inside the linesearch.
This might not be the optimal solution. For example, as `lin_fn` doesn't exist in `AbstractGaussNewton.step`, I just passed the identity function there. Let me know if you know a better way to handle this.

To avoid recalculating a gradient in the `accepted` branch of `AbstractQuasiNewton.step`, I added a check to read it from the search state if it's using zoom linesearch, which might not be the most elegant solution.

I haven't added any new tests yet, but I updated `BFGS` and `DFP` to use zoom instead of backtracking, and all tests pass for me.
I've also tested it with the L-BFGS implementation from #135. After that's merged, `LBFGS` should also be adopted to use `Zoom`.

I added, then removed a `StrongWolfeBacktracking` class that does backtracking and checks the strong Wolfe conditions. I can add it again or open a separate PR if that's something you want to have.

### DIFF
--- a/docs/api/searches/searches.md
+++ b/docs/api/searches/searches.md
@@ -32,3 +32,10 @@
     options:
         members:
             - __init__
+
+---
+
+::: optimistix.Zoom
+    options:
+        members:
+            - __init__


johannahaffner:
Nice! I'm just going through the L-BFGS PR in https://github.com/patrick-kidger/optimistix/pull/135. 

Since this is a larger change, shall we schedule a Zoom discussion for next week? I think the strong Wolfe would be nice to take a look at. If you still have it around, just put it in a separate draft PR as-is and I'll take a look :)

bagibence:
Sure, let's do that! I'll write you an email.

optimistix/_solver/quasi_newton.py
```diff
@@ -556,7 +578,7 @@ class DFP(AbstractQuasiNewton[Y, Aux, _Hessian]):
     atol: float
     norm: Callable[[PyTree], Scalar]
     descent: NewtonDescent
-    search: BacktrackingArmijo
+    search: AbstractSearch
```
johannahaffner: We intentionally provide concrete classes in the concrete solvers :)

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,951 @@
...
+            point_middle.value,
+            slope_middle,
+            state.init_point.value,
+            state.slope_init,
+        )
+        middle_satisf_curvature = self.curvature_condition(
+            slope_middle, state.slope_init
+        )
+
+        if self.verbose:
```
johannahaffner: We have our own `verbose_print`, take a look at how this is handled in `AbstractQuasiNewton`.

johannahaffner (commented):
Two very minor comments, not very important in the grand scheme of things. I took a look through, a bit more high-level for now.

Let's walk through this tomorrow, starting from the math and really looking at what is required and where. I'm also going to take a look at the optax and JAXopt implementations from which you are adapting things. 

In particular we will need solutions for: 

- communicating the first direction (`y - y_eval` after an accepted step) to the search; 
- getting gradients at the trial iterate in a way that a) does not compromise the performance of existing solvers which depends on skipping this, and b) does not introduce hard-coded coupling to specific search flavours in the solvers (if it can be avoided)
- how to avoid passing a `lin_fn` to the search, since this is a solver-level implementation detail (as you have seen the least squares solvers do not provide this)
- figuring out if this needs to be a two-loop setup / if this is a good fit for the existing "flattened" loop architecture

We could very well be looking at a break to the existing logic here, so this will need some thinking :)

optimistix/_solver/quasi_newton.py
bagibence: Okay. Do you want me to revert this back or switch to `Zoom`? Also, should I remove `search` from `__init__`?

johannahaffner:
> I got rid of the `_FnEvalInfo` altogether and switched to `FunctionInfo.EvalGrad` everywhere. 

I think that is clear!

> Would you prefer `ValueError` instead?

Yes, like it is done here: https://github.com/patrick-kidger/optimistix/blob/9927984fb8cbec77f9514fad7af076dce64e3993/optimistix/_solver/gauss_newton.py#L72

>  However, since it's ultimately determined by the search used if it needs the gradient at `y_eval` or not, I added a `_needs_grad_at_y_eval` class attribute to the search classes instead. 

This is a very good point and I do like this solution! 

> I agree that it would be nice to have that, and I made the `PointEval` and `PointEvalGrad` classes for that convenience. But wouldn't `FunctionInfo.at` make passing around both `y` and `f_info` (and `y_eval` and `f_eval_info`) redundant in many places throughout the codebase?
> [Plus updates 1, 2, 3]

And these are excellent points too! Then let's hold off on mainstreaming this change - I think the wrapper class approach makes sense, and could potentially even be extended to include other point-specific information, such as the slope?

johannahaffner:
Just rebased on latest dev, will take a long look at it tomorrow! Sorry for being so slow on this, will get the ball rolling now :)

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+                "In other words, variable `f_eval_info` needs to be"
+                "of type `FunctionInfo.EvalGrad`."
+            )
+
+        _fake_first_step_fn = ft.partial(
+            self.fake_first_step, y, y_eval, f_info, f_eval_info
+        )
+        _step_fn = ft.partial(self._step, y, y_eval, f_info, f_eval_info)
+
+        accept, state = jax.lax.cond(
```
johannahaffner, line 898: note for later - we can use `filter_cond` from the misc module in case we want to put a non-array into the `ZoomState`.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+                ~accept & state.failed & (state.safe_stepsize <= 0.0),
+                "Failed and safe stepsize is invalid.",
+            )
+            _cond_print(
+                ~accept & state.failed & (state.safe_stepsize > 0.0),
+                "Failed but proposing safe stepsize.",
+            )
+
+        # write these into the state
+        state = eqx.tree_at(
```
johannahaffner, line 952: If I'm not mistaken, then we modify the state in various places using a `tree_at`, which makes it hard to follow what happens where. Could that be changed? It probably requires breaking up this large state into several sub-states, e.g. for the interpolation? I think it would be very desirable to update the state in one place only.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+    increase_factor: float = 1.5
+    initial_guess_strategy: str = "keep"
+    min_interval_length: float = 1e-6
+    min_stepsize: float = 1e-6
+    maxls: int = 30
+    verbose: frozenset[str] = frozenset()
+    _needs_grad_at_y_eval: ClassVar[bool] = True
+
+    @staticmethod
+    def _replace_step_with_safe(state: ZoomState):
```
johannahaffner: This method can go into the main step function, see my other comment on unifying updates to the search state.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+        - `accepted`: Print when accepting a stepsize.
+        - `failed`: Print if the linesearch failed and if the safe stepsize is valid.
+        - `linesearch_diagnostics`: Print deeper diagnostics such as which Wolfe
+        conditions are satisfied, the interval we're zooming into, if the interval
+        is too short, etc.
+        - `result_per_ls_iter`: Print linesearch iteration, the checked stepsize,
+        and if we're done or failed.
+    """
+
+    # TODO decide on defaults
```
johannahaffner, line 227: I believe that we should always satisfy 0 < c1 < c2 < 1, so it would be good to check for this condition here. Take a look at `BacktrackingArmijo`, which checks parameter validity with a `__post_init__`.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+        """
+        del f_info_struct
+
+        _slope_init = jnp.array(-jnp.inf)
+        init_point = PointEvalGrad(
+            y, FunctionInfo.EvalGrad(jnp.array(jnp.inf), tree_full_like(y, jnp.inf))
+        )
+
+        return ZoomState(
+            ls_iter_num=jnp.array(0),
```
johannahaffner, line 379: Do we care about the number of iterations in the increase/zoom stage or do we only care about the outermost of these conditions, namely whether we are about to start a new linesearch, having successfully completed the last one? In the latter case, does this contain the same information as `done` and `failed`?

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,962 @@
...
+            init_point=init_point,
+            slope_init=_slope_init,
+            #
+            stepsize=jnp.array(0.0),
+            current_point=init_point,
+            current_slope=_slope_init,
+            #
+            interval_found=jnp.array(False),
+            done=jnp.array(False),
```
johannahaffner, line 388: Just to confirm, do `interval_found`, `done` and `failed` encode completely different elements of the state? As in, they do not have a logical relationship that would allow us to express one of these through the others?

optimistix/_solver/zoom.py
bagibence: We need to keep track of the number of iterations for checking if we reached the max number of steps allowed (`maxls`).

optimistix/_solver/zoom.py
bagibence: They are different.

`interval_found` is separate, it's for switching between the stages of the algorithm: looking for an interval and "zooming into it".

`done` and `failed` are related, but not as simply as `done = ~failed`, they kind of work together:

`not done, not failed` - continue.
`done, not failed` - accept the currently tested stepsize.
`not done, failed` - try the safe stepsize on the next iteration.

If needed, we might be able to eliminate `state.done` if `_search_interval` and `_zoom_into_interval` return `accept, state` instead of just the state.

optimistix/_solver/zoom.py
bagibence: I added the checks in `Zoom.__post_init__` and a short test.

optimistix/_solver/zoom.py
bagibence: I removed this method and put it in `step`, removing one `tree_at`.

optimistix/_solver/zoom.py
bagibence: I think I see what you mean.
Now I reduced it to a single `tree_at`. I'll think about how we could get rid of that. Or is that fine?

optimistix/_solver/zoom.py
johannahaffner: Great!

optimistix/_solver/zoom.py
```diff
@@ -221,7 +221,7 @@ class Zoom(AbstractSearch[Y, _FnInfo, FunctionInfo.EvalGrad, ZoomState]):
     # TODO decide on defaults
     c1: float = 1e-4
     c2: float = 0.9
-    c3: float = 1e-6
+    c3: float | None = 1e-6
```
johannahaffner, line 230: Can we document why `c3` is optional and what the search does if it is not specified?

optimistix/_solver/zoom.py
johannahaffner: Nothing wrong with a `tree_at` in principle! It makes it a little harder to see what does not get updated, though. Do we have elements of the search state that are unchanged here? Intuitively I would assume that the search state is updated at every `search.step`, even if that means intentionally _not_ updating certain elements of it. (Pretty much like `solver.state` is also updated in every `solver.step`, even though some elements may emerge unchanged when we take the `rejected` branch.)

optimistix/_solver/zoom.py
johannahaffner: Thank you for the explanation! Let's keep it for now, I will try to organise my thoughts with pen and paper.

johannahaffner:
I'm making dents in it, I promise 🙈 can I rebase this on a tidy development branch?

optimistix/_solver/zoom.py
bagibence: I added `Zoom.__init__.__doc__` with an explanation for this and comments in `Zoom.decrease_condition_with_approx`.
Let me know if you want to add a more exhausting documentation somewhere.

The [Optax docs](https://optax.readthedocs.io/en/latest/api/transformations.html#optax.scale_by_zoom_linesearch) are really good, although a bit confusing because if I understand things correctly, they don't necessarily switch to the approximate condition but allow either of Armijo or the approximate one in case the function doesn't change much.

johannahaffner:
Ok, done! I also squashed your work on the Zoom line search into a single commit (attributed to you, of course)! The tests now pass and we're up to date with the latest version of `tidy-dev`. You'll need to pull from this, or perhaps reset your local copy on this version. 

Sorry for the confusion! This is not really part of the PR experience I want to offer. Now sinking my teeth into the math of this :)

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+def optax_cubicmin(
+    a: FloatScalar,
+    value_a: FloatScalar,
+    slope_a: FloatScalar,
+    b: FloatScalar,
+    value_b: FloatScalar,
+    c: FloatScalar,
+    value_c: FloatScalar,
+) -> FloatScalar:
+    """
```
johannahaffner: Nit: text in docstrings starts in same line, e.g. 

```python
"""A docstring.
With extra text.
"""

File: optimistix/_solver/zoom.py [hunk 11]
```diff
@@ -0,0 +1,1017 @@
...
+
+    Finds a critical point of a cubic polynomial
+    p(x) = A *(x-a)^3 + B*(x-a)^2 + C*(x-a) + D, that goes through the
+    points (a,value_a), (b,value_b), and (c,value_c) with derivative at a of slope_a.
+    May return NaN (if radical<0), in that case, the point will be ignored.
+    Adapted from scipy.optimize._linesearch.py.
+
+    Args:
+      a: scalar
+      value_a: value of a function f at a
```
johannahaffner: `a`, `b`, `c` and the associated values are step lengths, right? And `f` is not the function to be minimised, but the step length selection function, where the step lengths are what is usually called $\alpha$ in the literature and `f` would usually be called $\phi$. 

Can we clarify that here? I think one way to do it would be to change the names here. Is there some ordering of the values `a`, `b` and `c`, e.g. `a < b < c`? Perhaps we can also call `f` something else, e.g. `steplength_fn`.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+      b: scalar
+      value_b: value of a function f at b
+      c: scalar
+      value_c: value of a function f at c
+
+    Returns:
+      xmin: point at which p'(xmin) = 0
+    """
+    C = slope_a
+    db = b - a
```
johannahaffner: If the names of `a` and `b` don't make it obvious, call this `distance_{...}` or `distance_to_{...}`, since are in relation to `a`.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+    lo: FloatScalar,
+    value_lo: FloatScalar,
+    slope_lo: FloatScalar,
+    hi: FloatScalar,
+    value_hi: FloatScalar,
+    cubic_ref: FloatScalar,
+    value_cubic_ref: FloatScalar,
+) -> FloatScalar:
+    """
+    Find a stepsize by minimizing the cubic or quadratic curve fitted to
```
johannahaffner: Nit: docstring, as above.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+    `lo`, `hi`, and `cubic_ref`.
+    """
+    # adapted from optax
+    delta = jnp.abs(hi - lo)
+    left = jnp.minimum(hi, lo)
+    right = jnp.maximum(hi, lo)
+    cubic_chk = 0.2 * delta  # cubic guess has to be at least this far from the sides
```
johannahaffner: How about `cubic_threshold` and `quadratic_threshold`?

optimistix/_solver/zoom.py
johannahaffner: This has been a source of confusion for me too. I'm also wondering if we should actually offer this feature in this search, as opposed to creating an abstract base class for Zoom line searches that then have slightly different features. 

We're doing something similar with the linear and classical trust region searches, for example. The concrete classes / searches that subclass the base class don't have to do a whole lot differently, but we can separate the functionality a little more cleanly, and document for users what is happening. 

Does that make sense to you? I think there are a few extra features here - same for the initial step length selection - and quite a bit of control flow already, making it tricky to understand what is supposed to happen if these different options interact.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+
+        return a_i
+
+    def _actual_init(
+        self,
+        init_point: PointEvalGrad,
+        y_eval_stepsize: FloatScalar,
+        y_eval: Y,
+    ) -> ZoomState:
+        """
```
johannahaffner: Nit: docstring :)

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+        Init function actually used when starting a new linesearch,
+        aka when the number of linesearch steps is reset to 0.
+
+        Instead of initializing the stepsize here, we use the stepsize that was proposed
+        at the end of the last linesearch step and was used to create the `y_eval` here.
+        """
+        # init_point is where stepsize = 0
+        descent_direction = (y_eval**ω - init_point.location**ω).ω
```
johannahaffner, line 327: Isn't `init_point` always `y`? In this case this would be `y_diff`, for consistency with the other searches. I understand that we have different layers of accepted things here (intervals, ...); but the overall solver logic for accepting `y` does not change, does it? So `y` would remain constant while only `y_eval` changes, and since this is a line search the descent direction is actually always the line that connects `y` with any `y_eval`.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+        descent_direction = jax.tree.map(
+            lambda x: x / y_eval_stepsize,
+            descent_direction,
+        )
+        _slope_init = init_point.compute_grad_dot(descent_direction)
```
johannahaffner, line 332: This would now be the slope of the univariate step length selection function, $\phi(\alpha)$?

optimistix/_solver/zoom.py
johannahaffner: Makes sense! Can we call it `line_search_max_steps`?

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+            safe_point=init_point,
+            safe_slope=_slope_init,
+            #
+            y_eval_stepsize=jnp.array(-jnp.inf),
+            #
+            descent_direction=tree_full_like(y, jnp.inf),
+        )
+
+    def _propose_by_interpolation(self, state: ZoomState) -> FloatScalar:
+        """
```
johannahaffner: docstring convention :)

johannahaffner: Does it make sense to have this wrapper around the interpolation methods? It seems a little onion-y, since this is a thin wrapper around two small functions.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+            [state.safe_stepsize, state.safe_point, state.safe_slope],
+        )
+
+        #
+        middle_slope_satisf_third_cond = (
+            slope_middle * (state.stepsize_hi - state.stepsize_lo) >= 0
+        )
+
+        # new point is not better than lo, so replace the hi side with it, keep lo as lo
+        set_hi_to_middle = (~middle_satisf_decrease) | (~middle_lower_than_lo)
```
johannahaffner, line 546: nit: can we just use `high` and `low`? We're saving only one and two characters with this abbreviation.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+        # otherwise hi stays hi
+        set_hi_to_lo = set_lo_to_middle & middle_slope_satisf_third_cond
+
+        # if set_hi_to_lo or set_hi_to_middle, then we overwrite hi
+        # and can use it as the reference point
+        # otherwise we changed lo, so keep that as reference
+        set_cubic_to_hi = set_hi_to_middle | set_hi_to_lo
+
+        # do the updates
+        new_stepsize_hi, new_point_hi = tree_where(
```
johannahaffner, line 564: I think it might be slightly more clear to update the step sizes first, and then use `new_point_high = (new_step_size_high * state.point_high**ω).ω`, instead of tree-mapping a tuple that we then unpack again immediately.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+        f_eval_info,
+        state: ZoomState,
+    ):
+        """Look for interval to zoom into."""
+        del y, f_info
+
+        # evaluate the slope along the descent direction for the new stepsize
+        new_stepsize = state.y_eval_stepsize
+        new_point = PointEvalGrad(y_eval, f_eval_info)
+        slope_at_new_point = new_point.compute_grad_dot(state.descent_direction)
```
johannahaffner, line 656: same as above, this is now the slope of the step size selection function $\phi$?

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+
+        # If the interval is found, from the next iteration on we do _zoom_into_interval
+        # if found_a: we call zoom(alpha_lo = alpha_{i-1}, alpha_hi = alpha_{i})
+        # if found_b: we call zoom(alpha_lo = alpha_{i}, alpha_hi = alpha_{i-1})
+        # where state.stepsize is alpha_{i-1} and new_stepsize is alpha_{i}
+
+        # If the interval is found, this will zoom into the correct interval.
+        # If not, it still sets lo and hi, but that's okay because it will not be used
+        # by _zoom_into_interval, and we will just return here in the next iteration.
+        new_stepsize_lo, new_point_lo, new_slope_lo = tree_where(
```
johannahaffner, line 702: See above, I think it would be preferable to update these separately instead of wrapping a tuple and then unpacking them again :)

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+            safe_stepsize=new_safe_stepsize,
+            safe_point=new_safe_point,
+            safe_slope=new_safe_slope,
+            #
+            y_eval_stepsize=state.y_eval_stepsize,
+            #
+            descent_direction=state.descent_direction,
+        )
+
+    def fake_first_step(
```
johannahaffner: Does this have to be a separate method?

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,1017 @@
...
+    ):
+        """
+        Only called once in the very beginning of the optimization.
+        Just accepts the proposed random point.
+        """
+        del y, y_eval, f_info, f_eval_info
+        accept = jnp.array(True)
+        return accept, state
+
+    def _safe_step(
```
johannahaffner: Same here, maybe just delineate that in `step` with a comment (what is now in the docstring)?

bagibence:
Hey! Sorry for the silence. I have the capacity to get back to this now and will start addressing your comments.

optimistix/_solver/zoom.py
bagibence: Okay, I changed all docstrings to use this convention.

optimistix/_solver/zoom.py
bagibence: Done.

optimistix/_solver/zoom.py
bagibence: Sure, I changed it.

optimistix/_solver/zoom.py
bagibence: Done.

optimistix/_solver/zoom.py
bagibence: Exactly, everything you write is correct.
`descent_direction` could be called `y_diff`, but I think this is clearer in this case. It's calculated once at initialization and normalized by `y_eval_stepsize`. For me `y_diff` is something you create for all `y_eval` points and is the actual difference.

johannahaffner:
The one failure during pre-commit is due to a newer pyright version, I think. We fixed something like that in https://github.com/patrick-kidger/equinox/pull/1100/files, although I have not looked at the error in any depth (gotta run!).

optimistix/_solver/zoom.py
bagibence: Yes, exactly. In other words the slope of the objective function along `descent_direction`.

optimistix/_solver/zoom.py
bagibence: I see what you mean. For me the branching was easier to follow and document by breaking it up like this.
How else would you prefer it? Define `_propose_by_interpolation` and `_propose_by_increase` inside `propose_stepsize`? Or move `interpolate`'s body into `_propose_by_interpolation`?

optimistix/_solver/zoom.py
bagibence: I think I did this so that line lengths work out and because they're the same length this way.
Also to signal to myself a bit that they are objects rather than adjectives / adverbs. If I see `low` or `high` I automatically assume that it's a bool.
If you prefer `high` and `low`, I'm happy to change it, no hard preferences.

optimistix/_solver/zoom.py
bagibence: Not necessarily. I pulled it out because it was easier to follow/document.

optimistix/_solver/zoom.py
bagibence: I could do that. Or have a method called something like `_just_accept` that replaces these two as they do the same thing?

optimistix/_solver/zoom.py
bagibence: Of course, renamed it.

optimistix/_solver/zoom.py
bagibence: This is just a helper that works for any univariate function that I just called `f` here for simplicity and 3 points on it.
But yes, when it's called during the search, `a`, `b`, `c` are all step lengths, and because they are all along the same line the function corresponds to $\phi$.

I don't think there is a strict ordering here. `stepsize_lo`, `stepsize_hi` can be left or right, and `cubic_ref` is generally outside the interval I think.

optimistix/_solver/zoom.py
bagibence: Yes, slopes are always along the descent direction, so derivatives of $\phi$.

optimistix/_solver/zoom.py
bagibence: I think I prefer a single class with a few options. Then if you want to tweak behavior a bit, you just pass a different parameter.
I don't see these change behavior so much that they are a separate method. Jumping between classes to figure out what they do differently is probably more confusing (at least for me) than just seeing a small list of options in one place.

In NeMoS for example, if users wanted to just change the initialization strategy, they just pass a hyperparameter that we forward to the `Zoom` constructor.
With subclasses we would have to create some factory based on user options.
Then if you ever rename these subclasses or reorganize them, that would break that factory.

optimistix/_solver/zoom.py
bagibence: Not sure I understand. So get `new_stepsize_high` first with `jnp.where`s, then recalculate the new points?
If so, that would require the extra calculation of something like `new_location_high = y + new_stepsize_high * state.descent_direction` and an extra evaluation of `f(new_location_high)`.
We have both of these stored in the "points", and we don't have access to `f` in `_zoom_into_interval`.

bagibence: Or just have a few `jnp.where` instead of `tree_where`?

optimistix/_solver/zoom.py
johannahaffner: We have lots of methods already :) It makes it a bit harder to follow what is going on in some places, I think. I would be in favour of reducing these to the ones we definitely need, and handling the rest e.g. with inner functions, where that helps make things more clear.

optimistix/_solver/zoom.py
johannahaffner: See my comment below, I think we have too many rather than not enough separate methods :) Makes for quite a lot of scrolling up and down to understand what is happening.

optimistix/_solver/zoom.py
johannahaffner: Ah I see! Then never mind :)

optimistix/_solver/zoom.py
johannahaffner: Let me get back to you on that! I think I'm confused about where we need the normalised descent direction and where we need the full step. 

In any case, `init_point` is exactly `y`, right? In that case it does not need another name.

optimistix/_solver/zoom.py
johannahaffner: I see! Then let's find a descriptive name for `f`, something that the unsuspecting reader will immediately understand.

optimistix/_solver/zoom.py
johannahaffner: I had some loose ideas on this, let me get back to you with those!

optimistix/_solver/zoom.py
johannahaffner: No strong feelings either :)

optimistix/_solver/zoom.py
bagibence: The normalized descent direction is only needed for calculating the directional derivative (= slope) along the line.

They're not exactly the same. `init_point` wraps `y` and `f_info`, so `y` is `init_point.location`

optimistix/_solver/zoom.py
bagibence: Okay! Then for now I'll leave it like this.

optimistix/_solver/zoom.py
bagibence: I removed `fake_first_step` and `_safe_step`.

optimistix/_solver/zoom.py
bagibence: I removed `fake_first_step` and `_safe_step`, replaced them with small inner functions.

optimistix/_solver/zoom.py
johannahaffner: So right now we have about 1k lines of code here. This is a lot :D 
It also means that we should think about how we can divide and conquer this, such that we end up with something obvious and easy to maintain. I don't think we're there yet, and I do have to invest quite a bit of time to understand what is what, which is part of why it took me quite long to get around to this (the other part being PhD life). 

I appreciate that not having all options in a single solver might complicate things downstream. Is that perhaps something you could work around with an internal function that converts user options into the associated Optimistix solver in NeMos?

bagibence: Of course, if it could simplify things for you here, I can work around that downstream.

optimistix/_solver/zoom.py
bagibence: A distance can't be negative, so I renamed them to `b_minus_a` and `c_minus_a` instead. Let me know if it's fine.

optimistix/_solver/zoom.py
bagibence: Rephrased the docstring to avoid having to give a name to the function.

optimistix/_solver/zoom.py
```diff
@@ -0,0 +1,993 @@
...
+            self.propose_stepsize(state),
+        )
+
+        # if we failed on the current iteration (so failed but didn't accept yet)
+        # and the safe stepsize is valid
+        # then try backpedaling to the safe step
+        # and propose that instead so that we take one more step with it
+        # which will be accepted by _safe_step
+        # on the next iteration
+        try_safe_stepsize = ~accept & state.failed & (state.safe_stepsize > 0.0)
```
johannahaffner: Why can `safe_stepsize` ever fall below zero? (Can it?)

cjchristopher:
Replying here as per request in https://github.com/patrick-kidger/optimistix/pull/119

> @cjchristopher did they have an L-BFGS-B with a Zoom line search? That sounds like an odd choice. 
> 
> PS: putting this here - happy to continue the conversation elsewhere, once we have a new open PR for any of these things :)

I'm not sure where to respond in the immediacy so I hope this doesn't do anything weird like reopen this PR @johannahaffner - in addition to issue 620 linked my last reply, to answer your question about the jaxopt L-BFGS-B: https://github.com/google/jaxopt/blob/cf28b4563f5ad9354b76433622dbb9ee32af5f09/jaxopt/_src/lbfgsb.py#L288 -  which has been set like that since the file was first committed to the repo - possibly an artefact of just copy-pasting from l-bfgs to get it started?

Having said that, if you go back to the original paper: https://dl.acm.org/doi/10.1145/279232.279236 - it references a line search by [Moré & Thuente](https://dl.acm.org/doi/10.1145/192115.192132)  - and digging up that algorithm, it looks very much *like* a zoom-type search (see https://github.com/antoinecollet5/lbfgsb/blob/master/lbfgsb/linesearch.py for a port to python, that uses scipy's reference implementation for dcsrch- and apparently scipy's entire L-BFGS-B uses this as well). So I'm not 100% sure on the whole yes-or-no to a Zoom for bounded/constrained non-convex opt. That jaxopt gave such divergent results for their l-bfgs-b affecting multiple people makes me think there has to be a problem somewhere.

Re Zoom stuff - Having spend an excruciating amount of time going over the detail with respect to Nocedal and Wright, I've convinced myself that perhaps the jaxopt version is correct (as is the version here) - I'm going to have to dig up the version of my code that was producing the issues and attempt to retrace where the errors were occurring - it's possible the issue is actually with the jaxopt L-BFGS-B somewhere, but that really disagrees with my memory.

PR #145, opened 2025-06-29
Anderson acceleration methods

floffy-f:
## New fixed point method: Anderson

Thank you very much for this amazing library.

I am attempting to implement the Anderson acceleration algorithm for fixed point systems, to contribute to solve #3.
So I would appreciate some advices on this PR, hoping that I did not miss any corner case.

### Description

**Goal**: solve $y^{\*}=f(y^{\*})$ with Anderson acceleration.

A step is performed by combining old residuals. We build a matrix $G$ containing columnwise the residuals at the `memory_size` last iterations.
We solve the minimization problem for $\alpha^{(k)}$:

$$
\begin{align}
  \min_{\alpha\in\mathbb{R}^d} & \|\|G\alpha^{(k)}\|\|_2^2\\
  \text{s.t.} & \sum_i \alpha_i^{(k)}
\end{align}
$$

This is equivalent to solve its KKT conditions system:

$$
\left[\begin{array}{cc}
  0 & \mathbb{1}^T\\
  \mathbb{1} & G^TG
\end{array}\right]
\left(\begin{array}{c}
  \nu^{(k)}\\
  \alpha^{(k)}
\end{array}\right) =
\left(\begin{array}{c}
  1\\
  \mathbb{0}
\end{array}\right)
$$

The user must thus provide a linear solver to compute $\alpha^{(k)}$.
Then we get as the update formula:

$$
y^{(k+1)} = \sum_{i\in\texttt{memory}} \alpha_i^{(k)}f(y^{(i)}).
$$

Adding some dampening to this dynamic looks like:

$$
y^{(k+1)} = (1-\beta)\sum_{i\in\texttt{memory}} \alpha_i^{(k)}y^{(i)} + \beta\sum_{i\in\texttt{memory}} \alpha_i^{(k)}f(y^{(i)}).
$$

So we keep a buffer with the past iterates $\left(y^{(i)}\right)_{i\in\texttt{memory}}$,

and residuals $\left(f(y^{(i)})-y^{(i)}\right)_{i\in\texttt{memory}}$.

---

### Some quick questions

1) Regarding the documentation, the `docs/requirements.txt` file is non-existant in the `dev` branch, I do not know if there is a good reason for that. Do I need to download it and push it in my PR?

2) In the linear solve for the anderson step, the `memory_size` first steps are not trivial to compute since the size of the system to solve changes... What I did is to make a "rank 1" memory with a `memory_size*dim` sized matrix of `memory_size` copies of the first iterate.
   My question is: could you think of another way of initializing the memory buffers? E.g.:
   - would initializing the residual buffer with all $+\infty$s, and the iterates buffer with $0$s be a good idea? Is `lineax` compatible with this somehow? In this case, the first iteration in the `init` function could be avoided by tracing with `jax.eval_shape`.
   - may I introduce `dim` new lines (and dual multipliers) to the system to force $(\alpha_i^{(k)})_{i\ge k}=0$?
This question comes from the fact that with what I implemented, the G^T G matrix at the few first iterations is ill-conditioned (rank k, with k the current iteration).

3) I am not sure of the best way to use the numpy context manager for dtype promotion. Is it costly to perform operations inside?

4) Do you know of some test case for which Anderson converges quickly while the fixed point solver should diverge? If you do, I may add it to the test suite to check my implementation.

---

### Some implementation details

I implemented the update formula with the dampening, with default "`mixing`" $\beta=\frac1{2}$, but also put a "`regularization`" on the system to help with the first few steps and last ones close to the solution (where $G\to 0$ as residuals get small), in the same spirit as Ridge for least square problems.

The solver is tested for $\beta=$`0.5` and $\beta=$`0.3`, in the same file as the fixed point solver, and for the same problems.

### DIFF
--- a/optimistix/_solver/__init__.py
+++ b/optimistix/_solver/__init__.py
@@ -1,3 +1,6 @@
+from .anderson_accelerated_fixedpoint import (
+    AndersonAcceleration as AndersonAcceleration,
+)
 from .backtracking import BacktrackingArmijo as BacktrackingArmijo
 from .best_so_far import (
     BestSoFarFixedPoint as BestSoFarFixedPoint,
--- a/optimistix/_solver/anderson_accelerated_fixedpoint.py
+++ b/optimistix/_solver/anderson_accelerated_fixedpoint.py
@@ -1,0 +1,254 @@
+from collections.abc import Callable
+from typing import Any, Generic
+
+import equinox as eqx
+import jax
+import jax.numpy as jnp
+import lineax as lx
+from equinox.internal import ω
+from jaxtyping import Array, Bool, Int, PyTree, Scalar
+
+from .._custom_types import Aux, Fn, Y
+from .._fixed_point import AbstractFixedPointSolver
+from .._misc import max_norm
+from .._solution import RESULTS
+
+
+class _AndersonAccelFixedPointState(eqx.Module, Generic[Y]):
+    relative_error: Scalar
+    past_errors: Y
+    past_iterates: Y
+    right_hand_side: tuple[Scalar, Array]
+    current_iter: Int[Array, ""]
+
+
+class AndersonAcceleration(
+    AbstractFixedPointSolver[Y, Aux, _AndersonAccelFixedPointState]
+):
+    r"""Repeatedly calls a function in search of a fixed point.
+
+    Unlike the usual fixed point iterations that simply
+    repeatedly call `y_{n+1}=f(y_n)` until `y_n` stops changing,
+    Anderson methods attempt to blend residuals (`f(y_k)-y_k`) and
+    iterates (`y_k`) from a window of past iterations to produce
+    the next iterate.
+
+    The update formula amounts to finding `α` solving the following
+    optimisation problem.
+
+    ```
+    min || G.α ||^2
+    s.t. 1^T.α = 1
+         G = [*residuals_past]
+    ```
+    The KKT conditions of this problem yield a system of linear equations to
+    solve at each iteration, of size `dim(y)+1`.
+    ```
+    [ 0  1^T  ] [ν]   [1]
+    [         ] [ ]   [ ]
+    [ 1 G^T.G ] [α] = [0]
+    [         ] [ ]   [ ]
+    ```
+    Then combine the past residuals and iterates:
+    ```
+    new_y = β * dot(G, α) + dot([*y_past], α)
+    ```
+
+    See e.g.
+        - [this tutorial](https://implicit-layers-tutorial.org/implicit_functions/).
+        - [[Fang and Saad, 2009]](https://onlinelibrary.wiley.com/doi/10.1002/nla.617)
+        - [[Walker and Ni, 2011]](https://users.wpi.edu/~walker/Papers/Walker-Ni,SINUM,V49,1715-1735.pdf)
+    """
+
+    rtol: float
+    atol: float
+    norm: Callable[[PyTree], Scalar] = max_norm
+    # Beta coefficient for damping
+    mixing: float = 0.5
+    memory_size: int = 10
+    # Regularization factor to lift up the smallest e.v. of the normal equations:
+    # Cf. doi.org/10.1016/j.anucene.2024.110837 eq (42)->(43)
+    regularization: float = 1e-6
+    linear_solver: lx.AbstractLinearSolver = lx.LU()
+
+    def init(
+        self,
+        fn: Fn[Y, Y, Aux],
+        y: Y,
+        args: PyTree,
+        options: dict[str, Any],
+        f_struct: PyTree[jax.ShapeDtypeStruct],
+        aux_struct: PyTree[jax.ShapeDtypeStruct],
+        tags: frozenset[object],
+    ) -> _AndersonAccelFixedPointState:
+        del options, f_struct, aux_struct, tags
+        fn_output, _ = fn(y, args)
+
+        with jax.numpy_dtype_promotion("standard"):
+            # The matrix involved in the normal equations (G)
+            error_memory_init = jax.tree.map(
+                lambda _f, _y: jnp.stack([_f - _y] * self.memory_size, axis=0),
+                fn_output,
+                y,
+            )
+            # [z_k, z_{k-1}, ...]
+            iterates_memory_init = jax.tree.map(
+                lambda _y: jnp.stack([_y] * self.memory_size, axis=0), y
+            )
+            # Get the kind of (nu, alpha) combination wished for, by computing
+            # the dtype of the output. _infos contains the dtype of a scalar that
+            # corresponds to the sum of all the coordinates (in pytree form) of
+            # the function's output
+            _infos = jax.eval_shape(
+                lambda _y: jax.tree.reduce(jnp.add, (_y**ω).call(jnp.sum).ω), fn_output
+            )
+        # Now that we used context management to get the "common dtype" of the
+        # outputs of the function, i.e. of the linear system to solve, we can
+        # build the right hand side of the normal equations
+        right_hand_side = (
+            jnp.ones_like(_infos),
+            jnp.zeros((self.memory_size,)).astype(_infos.dtype),
+        )
+        return _AndersonAccelFixedPointState(
+            jnp.array(jnp.inf),
+            error_memory_init,
+            iterates_memory_init,
+            right_hand_side,
+            jnp.array(1),
+        )
+
+    def step(
+        self,
+        fn: Fn[Y, Y, Aux],
+        y: Y,
+        args: PyTree,
+        options: dict[str, Any],
+        state: _AndersonAccelFixedPointState,
+        tags: frozenset[object],
+    ) -> tuple[Y, _AndersonAccelFixedPointState, Aux]:
+        del tags
+        fn_output, aux = fn(y, args)
+        error = y**ω - fn_output**ω
+
+        # Update the rolling memory without shifting the data
+        new_errors_memory: Y = jax.tree.map(
+            lambda olds_, new_: olds_.at[state.current_iter % self.memory_size].set(
+                new_
+            ),
+            state.past_errors,
+            error.ω,
+        )
+        new_iterates_memory: Y = jax.tree.map(
+            lambda olds_, new_: olds_.at[state.current_iter % self.memory_size].set(
+                new_
+            ),
+            state.past_iterates,
+            fn_output,
+        )
+        with jax.numpy_dtype_promotion("standard"):
+
+            def kkt_linear_op(
+                optim_vector: tuple[Scalar, Array],
+            ) -> tuple[Scalar, Array]:
+                """
+                Solve:
+                [ 0 1...1 ] ( ν )   ( 1 )
+                | 1       | |   |   | 0 |
+                | : G^T.G | | α | = | : |
+                [ 1       ] (   )   ( 0 )
+                This function describes the left operator.
+                """
+                nu, alpha = optim_vector
+                normal_equation = nu + jax.tree.reduce(
+                    lambda _carry, _errs: _carry
+                    + jnp.einsum("j...,i...,i->j", _errs, _errs, alpha),
+                    new_errors_memory,
+                    initializer=jnp.array(0.0),
+                )
+                return (jnp.sum(alpha), normal_equation + self.regularization * alpha)
+
+            # Pointer to the right-hand side of the equations
+            rhs = state.right_hand_side
+
+            solution = lx.linear_solve(
+                # In case the user wishes to use GMRES/NormalCG as solver,
+                # we define the linear operator lazily without materializing the
+                # matrix
+                lx.FunctionLinearOperator(
+                    kkt_linear_op,
+                    jax.eval_shape(lambda: rhs),
+                    # The normal equations matrix is always at least symmetric.
+                    # It has one negative eigenvalue, and n positive ones.
+                    tags=frozenset({lx.symmetric_tag}),
+                ),
+                rhs,
+                self.linear_solver,
+                throw=False,
+                options=options,
+            )
+            _, weighting = solution.value
+
+            # New iterate is G.α
+            new_y = jax.tree.map(
+                eqx.Partial(self._combine_iters, weighting),
+                new_errors_memory,
+                new_iterates_memory,
+            )
+
+            scale = self.atol + self.rtol * ω(new_y).call(jnp.abs)
+            new_state = _AndersonAccelFixedPointState(
+                self.norm((error / scale).ω),
+                new_errors_memory,
+                new_iterates_memory,
+                state.right_hand_side,
+                state.current_iter + 1,
+            )
+        return new_y, new_state, aux
+
+    def _combine_iters(
+        self, weighting: Array, errors: PyTree[Array], iters: PyTree[Array]
+    ):
+        # Helper function to combine past iterates and past residuals into next
+        # iterates.
+        return jnp.einsum("i...,i->...", iters + self.mixing * errors, weighting)
+
+    def terminate(
+        self,
+        fn: Fn[Y, Y, Aux],
+        y: Y,
+        args: PyTree,
+        options: dict[str, Any],
+        state: _AndersonAccelFixedPointState,
+        tags: frozenset[object],
+    ) -> tuple[Bool[Array, ""], RESULTS]:
+        return state.relative_error < 1, RESULTS.successful
+
+    def postprocess(
+        self,
+        fn: Fn[Y, Y, Aux],
+        y: Y,
+        aux: Aux,
+        args: PyTree,
+        options: dict[str, Any],
+        state: _AndersonAccelFixedPointState,
+        tags: frozenset[object],
+        result: RESULTS,
+    ) -> tuple[Y, Aux, dict[str, Any]]:
+        return y, aux, {}
+
+
+AndersonAcceleration.__init__.__doc__ = """**Arguments:**
+
+- `rtol`: Relative tolerance for terminating the solve.
+- `atol`: Absolute tolerance for terminating the solve.
+- `norm`: The norm used to determine the difference between two iterates in the 
+    convergence criteria. Should be any function `PyTree -> Scalar`. Optimistix
+    includes three built-in norms: [`optimistix.max_norm`][],
+    [`optimistix.rms_norm`][], and [`optimistix.two_norm`][].
+- `mixing`: Beta coefficient for damping. Defaults to `1.`
+- `memory_size`: Number of vectors to keep in memory, both for past residuals
+    and past iterates. Defaults to `10`
+- `regularization: regularization factor to lift up the smallest e.v. of the
+    normal equations. Defaults to `1e-6`
+- `linear_solver: Solver from the [`lineax`][] library. Defaults to [`lx.LU()`][]
+"""
--- a/tests/test_fixed_point.py
+++ b/tests/test_fixed_point.py
@@ -5,6 +5,7 @@ import equinox as eqx
 import jax.numpy as jnp
 import jax.random as jr
 import jax.tree_util as jtu
+import lineax as lx
 import optimistix as optx
 import pytest
 from equinox.internal import ω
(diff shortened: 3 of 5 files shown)


johannahaffner:
Thanks for the PR! I took a first look and will get back to you with feedback in the coming days :)

johannahaffner:
Sorry, that stretched my definition of a few days to a week! 

First off, the **quick questions**: 

> 1. Regarding the documentation, the `docs/requirements.txt` file is non-existant in the `dev` branch, I do not know if there is a good reason for that. Do I need to download it and push it in my PR?

We switched installing with `pip install -e '.[docs]'`, this is updated in `CONTRIBUTING.md` on `dev`, but not on `main`. 

> 2. In the linear solve for the anderson step, the `memory_size` first steps are not trivial to compute since the size of the system to solve changes... What I did is to make a "rank 1" memory with a `memory_size*dim` sized matrix of `memory_size` copies of the first iterate.

So over in https://github.com/patrick-kidger/optimistix/pull/135/ we solved this same issue with initialising using a pytree of all zeros, generated like this

```python
def _batched_tree_zeros_like(y, batch_dimension):
    return jtu.tree_map(lambda y: jnp.zeros((batch_dimension, *y.shape)), y)
```
Lineax will generally complain for infinite inputs, but not zeros. But these can affect the result of the linear solve, and what the right thing to do is depends on the nature of the linear system and affects the choice of solver (I'm going to write more about that in a bit). 

> In this case, the first iteration in the `init` function could be avoided by tracing with `jax.eval_shape`. 

This is a good idea, and happens to be what we do elsewhere too. 

>    * may I introduce `dim` new lines (and dual multipliers) to the system

Of course, I think this is the correct way to handle this here. I do something similar in KKT systems for active set algorithms, happy to point you to some code if you think it would be helpful!

> 3. I am not sure of the best way to use the numpy context manager for dtype promotion. Is it costly to perform operations inside?

Do we need that? I try to avoid this where possible. 

> 4. Do you know of some test case for which Anderson converges quickly while the fixed point solver should diverge? If you do, I may add it to the test suite to check my implementation.

This I can look up, there may be benchmark problems!

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+
+    ```
+    min || G.α ||^2
+    s.t. 1^T.α = 1
+         G = [*residuals_past]
+    ```
+    The KKT conditions of this problem yield a system of linear equations to
+    solve at each iteration, of size `dim(y)+1`.
+    ```
+    [ 0  1^T  ] [ν]   [1]
```
johannahaffner, line 47: Is this system correct? I'm a little confused by the empty lines, and my mind wants to interpret these as zeros. In which case the `1^T` actually gets multiplied by zero and maps to zero?

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+    rtol: float
+    atol: float
+    norm: Callable[[PyTree], Scalar] = max_norm
+    # Beta coefficient for damping
+    mixing: float = 0.5
+    memory_size: int = 10
+    # Regularization factor to lift up the smallest e.v. of the normal equations:
+    # Cf. doi.org/10.1016/j.anucene.2024.110837 eq (42)->(43)
+    regularization: float = 1e-6
+    linear_solver: lx.AbstractLinearSolver = lx.LU()
```
johannahaffner, line 72: `LU` can silently misbehave for singular systems in JAX, are we sure that we never get those?

I thought this was documented by https://github.com/jax-ml/jax/issues/26180, but apparently not - depending on how the system is conditioned, a reasonable-looking solution may be returned that won't be caught by checking for `NaN` values. Looks like I have to look into this again :)

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+        args: PyTree,
+        options: dict[str, Any],
+        f_struct: PyTree[jax.ShapeDtypeStruct],
+        aux_struct: PyTree[jax.ShapeDtypeStruct],
+        tags: frozenset[object],
+    ) -> _AndersonAccelFixedPointState:
+        del options, f_struct, aux_struct, tags
+        fn_output, _ = fn(y, args)
+
+        with jax.numpy_dtype_promotion("standard"):
```
johannahaffner, line 87: Mentioned before - why is this one required? This is not immediately obvious to me.

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+        options: dict[str, Any],
+        state: _AndersonAccelFixedPointState,
+        tags: frozenset[object],
+    ) -> tuple[Y, _AndersonAccelFixedPointState, Aux]:
+        del tags
+        fn_output, aux = fn(y, args)
+        error = y**ω - fn_output**ω
+
+        # Update the rolling memory without shifting the data
+        new_errors_memory: Y = jax.tree.map(
```
johannahaffner, line 134: Nit: he state already includes the type hints, so we don't need extra ones in the code.

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+            fn_output,
+        )
+        with jax.numpy_dtype_promotion("standard"):
+
+            def kkt_linear_op(
+                optim_vector: tuple[Scalar, Array],
+            ) -> tuple[Scalar, Array]:
+                """
+                Solve:
+                [ 0 1...1 ] ( ν )   ( 1 )
```
johannahaffner, line 155: Ah now this system of equations is clearer to me! Can we unify the variable names so that we use the same thing in the code and in the documentation? I'm not sure what $\nu$ is.

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+
+            solution = lx.linear_solve(
+                # In case the user wishes to use GMRES/NormalCG as solver,
+                # we define the linear operator lazily without materializing the
+                # matrix
+                lx.FunctionLinearOperator(
+                    kkt_linear_op,
+                    jax.eval_shape(lambda: rhs),
+                    # The normal equations matrix is always at least symmetric.
+                    # It has one negative eigenvalue, and n positive ones.
```
johannahaffner, line 181: The presence of positive and negative eigenvalues makes the normal equations system indefinite, are we sure that they cannot be singular?

optimistix/_solver/anderson_accelerated_fixedpoint.py
```diff
@@ -0,0 +1,254 @@
...
+
+AndersonAcceleration.__init__.__doc__ = """**Arguments:**
+
+- `rtol`: Relative tolerance for terminating the solve.
+- `atol`: Absolute tolerance for terminating the solve.
+- `norm`: The norm used to determine the difference between two iterates in the 
+    convergence criteria. Should be any function `PyTree -> Scalar`. Optimistix
+    includes three built-in norms: [`optimistix.max_norm`][],
+    [`optimistix.rms_norm`][], and [`optimistix.two_norm`][].
+- `mixing`: Beta coefficient for damping. Defaults to `1.`
```
johannahaffner, line 248: I think here (and above) it would be clearer to just explain what the mixing coefficient does. This goes into the documentation and will be read by people that don't know what $\beta$ usually refers to. We can instead state which of our variable names maps to what conventionally used symbol in the solver documentation, or even in a function documentation.

johannahaffner: I'm thinking about something like 

```python
- `mixing_coefficient`: a coefficient specifying the weight of the (linear combination) of
    past parameters `y` and past function values `f(y, args)` used when computing the 
    next value of `y`.
```

Feel free to adjust as needed if I'm getting this wrong!

Also I noticed that [here](https://en.wikipedia.org/wiki/Anderson_acceleration#Relaxation) they only mention that this coefficient must be non-negative, but I think from the formula it is clear that this must take a value between 0 and 1?

tests/test_fixed_point.py
```diff
@@ -20,7 +21,13 @@
 
 
 atol = rtol = 1e-6
-_fp_solvers = (optx.FixedPointIteration(rtol, atol),)
```
johannahaffner, line 23: This could go into `helpers`, and just become a non-private `fixed_point_solvers` - I'm going to use it to test vmapping stuff in another module, and the other solvers live there too.

optimistix/_solver/__init__.py
```diff
@@ -1,3 +1,6 @@
+from .anderson_accelerated_fixedpoint import (
```
johannahaffner, line 1: Ultranit: I think just `anderson.py` is enough :)

benchmarks/timing_anderson.py
```diff
@@ -0,0 +1,62 @@
...
+    stats = solve(y0)
+    dt = default_timer() - t0
+
+    stps = stats.mean().item()
+    return dt / stps, dt, stps
+
+
+if __name__ == "__main__":
+    key = jax.random.key(666)
+    print("FixedPointIteration:", time_solver(key, None))
```
johannahaffner, line 56: Do you also verify the correctness of the solutions returned? Iterative linear solvers usually require positive definite, symmetric matrices to work well.

johannahaffner (commented):
Alright, I've left a first round of comments! Nicely readable code, thank you!

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: Indeed with the white lines the system might look very unclear, my bad. I will rewrite it as above on the PR, but without the weird spacing like this if it looks better to you:
```python
'''
[ 0     1     ] [ν] = [1]
[ 1 G^TG ] [α] = [0]
'''
```
(my first goal was to show visually that ν is a scalar, and α a vector, but on a python comment without markdown it's sketchy... do you know by any chance if the version of sphinx formatting you use lets me use `ReStructuredText`'s tags like e.g. `..math::` to do latex instead by any mean?)

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: I copy-pasted nastily from the [`FixedPointIteration`](https://github.com/floffy-f/optimistix/blob/47e7277ebda6d8c054437db79dfdab6a4c4dc768/optimistix/_solver/fixed_point.py#L20) class (to handle complex numbers I guess?). I don't know where to use the context manager, and where not to, so I put it on the whole function... Do you have advice on how to transpose the code from `FixedPointIteration` properly? I must admit that I am not proficient with numpy/jax's casting mechanisms

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: Legit ! But the variable names are the same here as on the class docstring though, as well as this PR's notations. Is it more the spacing shenanigans that felt unclear, or the math notations? I'll take your advice on this, since I've had my head buried in Anderson for some time now so notations are a bit blurred for me, and we don't want documentation readers to feel that way I guess.

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: They actually can, indeed, which is the main problem I encountered during implem.
So what I did is to regularize the system by shifting all the positive e.v. up by some small coefficient. This is done in quite the same way you would do it in Ridge regression versus your usual least-squares, by adding a small perturbation times the identity matrix to the $G^TG$ matrix (i.e. the normal equations systems, **but not the dual multiplier**), see [l.168 of the file](https://github.com/floffy-f/optimistix/blob/47e7277ebda6d8c054437db79dfdab6a4c4dc768/optimistix/_solver/anderson_accelerated_fixedpoint.py#L168).

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: Fully agreed, I'll try to reformulate close to what you did there.
About the available mixing coefficient, fun fact is that $\beta$ does not need to be smaller than one. In the literature, $\beta>1$ is called "over-projection" or something like that. So yeah, that exists, and should work even better in some specific cases.
Do you think I should mention it in the doc? Also should I put a hard-check assertion to test out the $\beta\le 0$ case?

tests/test_fixed_point.py
floffy-f: I don't get what you meant there sorry. Do you mean that I should move the list of "test-cases" FP-solvers to this `helpers.py` file and put it in capslock as a constant that you wish to reuse elsewhere?

benchmarks/timing_anderson.py
floffy-f: Not at all, I can do this. I'll put this to lower priority, but I'll try to check it if I have some time.

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: Oh and about nu, it's the dual multiplier associated with the (scalar) constraint $\sum_i \alpha_i=1$. Confer my general answer for questions about the importance (or lack thereof) of this constraint

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: I did not check that. Indeed putting a safer option as default solver, say SVD, should be sounder.

floffy-f:
> Sorry, that stretched my definition of a few days to a week!

No problem!


> We switched installing with `pip install -e '.[docs]'`, this is updated in `CONTRIBUTING.md` on `dev`, but not on `main`.

Perfect, I'll try that.


> So over in #135 we solved this same issue with initialising using a pytree of all zeros, generated like this
> 
> ```python
> def _batched_tree_zeros_like(y, batch_dimension):
>     return jtu.tree_map(lambda y: jnp.zeros((batch_dimension, *y.shape)), y)
> ```
> 
> Lineax will generally complain for infinite inputs, but not zeros. But these can affect the result of the linear solve, and what the right thing to do is depends on the nature of the linear system and affects the choice of solver.

As discussed above, if the $G^TG$ matrix is full of zeros, then the full KKT system is rank two (hence non-invertible). This is a problem that we have by default near the optimal solution, but having it directly at the start of the algorithm is quite a pain I believe... So I would argue towards finding another way to initialize the matrix $G$ if any exists.

> > In this case, the first iteration in the `init` function could be avoided by tracing with `jax.eval_shape`.
> 
> This is a good idea, and happens to be what we do elsewhere too.

Perfect, I'll try to implement that then.

> Of course, I think this is the correct way to handle this here. I do something similar in KKT systems for active set algorithms, happy to point you to some code if you think it would be helpful!

Oh I see. I'll take a look at the code for active sets then, and in the meantime if you have any pointers I'd be glad to take them.


Thank you again for the time you took to look at my code and the useful advices. I'll get back to the PR soon to implement what we discussed, in light of your next answers.

johannahaffner:
For the linear least squares approach - this should work, the regularisation should force the terms to zero. 

Suppose we have $A x = b$, where $A$ is tall - then $x$ is a least squares solution. We can regularise this non-square operator by instead solving for $[A, \rho \cdot \mathbb{I}]^T x = [b, 0]^T$. This gives us a regularised operator, and as long as entire columns of $A$ are zero (after we initialised them), these get mapped to the zeros on the right-hand side. This means that for these elements $x_i$, we have that $\rho \cdot x_i = 0$, for some strictly positive $\rho$.

I think regularising in this way should be enough to get good properties with a least-squares solver such as `QR`, this is what we also do in the damped Newton step, e.g. here

https://github.com/patrick-kidger/optimistix/blob/9927984fb8cbec77f9514fad7af076dce64e3993/optimistix/_solver/levenberg_marquardt.py#L34

To illustrate this, here is a brief example with toy parameters:

```python
import jax.random as jr
import lineax as lx


residuals = jr.normal(jr.key(0), (6, 2))
zeros = jnp.zeros((6, 3))
values = jnp.concatenate((residuals, zeros), axis=1)

operator = lx.MatrixLinearOperator(values)
vector = jnp.array([1, 2, 0, 0, 0, 0])

unmodified_solution = lx.linear_solve(operator, vector, lx.SVD()).value

print(unmodified_solution)  # Does the right thing - but isn't forced to do so

regulariser = jnp.eye(5) * 1e-6

regularised_values = jnp.concatenate((values, regulariser), axis=0)
regularised_operator = lx.MatrixLinearOperator(regularised_values)
regularised_vector = jnp.concatenate((vector, jnp.zeros(5)), axis=0)

regularised_solution = lx.linear_solve(
    regularised_operator, regularised_vector, lx.SVD()
).value  # Also does the right thing (now needs to)
```

Now for the Lagrangian multipliers to ensure that the elements of $\alpha$ sum to one - I think here we can actually take any least squares solution we get and just normalise the values + adjust for scaling. This seems to be the inspiration behind [these approaches](https://en.wikipedia.org/wiki/Anderson_acceleration#Solution_of_the_minimization_problem) too.

floffy-f:
Indeed in [Walker et al.2018] they explain that there is an unconstrained equivalent formulation.
Using the same notations as above, it writes:
$$\min_\alpha \|\|G\alpha - r_k 1^T\alpha + r_k\|\|_2^2$$

If it is correct, then indeed it seems much simpler than the constrained one (with the KKT and everything), since it is directly written as a least squares problem. So that leaves SVD and QR as good options to solve the rectangular system directly.

And indeed, adding regularization as new lines to the tall matrix should work quite well. I am just worried that, as compared to adding $\texttt{regularizer}.\|\|\alpha\|\|_2^2$, it may incur a higher memory cost since the operator will be materialized by QR and SVD (does it?).

tests/test_fixed_point.py
johannahaffner: Then never mind! I'll make the change in another PR :)

optimistix/_solver/anderson_accelerated_fixedpoint.py
johannahaffner: Interesting! Definitely a bit counter-intuitive that we could have, for instance, $\beta = 6$ and then we add six times the first term to minus five times the second. 

I think the documentation should state that $\beta > 0$ (or $\beta \geq 0$?) is expected. 
If we want to go further, we could also check this with a `__post_init__` or a `__check_init__`, similar to what we do here: 
https://github.com/patrick-kidger/optimistix/blob/9927984fb8cbec77f9514fad7af076dce64e3993/optimistix/_solver/backtracking.py#L36

optimistix/_solver/anderson_accelerated_fixedpoint.py
johannahaffner: I checked this, and it is indeed necessary to handle complex numbers correctly!

optimistix/_solver/anderson_accelerated_fixedpoint.py
johannahaffner: Yes. Or we could use `lx.AutoLinearSolver(well_posed=None)`, which would default to `QR` for a rectangular matrix. (When switching to the linear least squares case.)

optimistix/_solver/anderson_accelerated_fixedpoint.py
johannahaffner: I don't think we're using `..math::` anywhere (and not sure if we would be compatible), but r-strings with long equations bracketed by `$` work, e.g. here: 

https://github.com/patrick-kidger/diffrax/blob/d3c1430b76e7d0158536823724371d7b26f5db8f/diffrax/_adjoint.py#L465

Haven't looked to deeply into this aspect of documentation!

optimistix/_solver/anderson_accelerated_fixedpoint.py
floffy-f: Good idea, I'll try that. And indeed this is counterintuitive, as Anderson himself advices to pick $\beta\in(0, 1\]$.
Note that if $\beta=0$ the algorithm might get stuck into a loop of the past iterations, and adding the residuals provides (according to [[Anderson,95]](https://dl.acm.org/doi/pdf/10.1145/321296.321305)) a "new dimension" for the search space.

johannahaffner: Then maybe let's cite Anderson with his suggestion of the parameter values, and mention in the docstring (with reference, if possible) that users may experiment with larger values, but we don't guarantee any behaviour for that? Something along those lines :)

optimistix/_solver/anderson_accelerated_fixedpoint.py
johannahaffner: Would you like some input on this? Happy to sit down and think up some names :)

PR #146, opened 2025-07-14
Inline methods for updates to the Hessian and inverse Hessian approximations

johannahaffner:
As discussed in https://github.com/patrick-kidger/optimistix/pull/135#discussion_r2188353615. 

Methods are now inlined, and the documentation has been updated. We inherit directly from the concrete classes in the helpers module for the tests, to avoid code duplication. This now breaks the tests for the LBFGS operators, these depended on (private) imports of classes that no longer exist. 

I think about how to resurrect the testing functionality for these, since the operators are complex and having specific tests for them would be good.

### DIFF
--- a/docs/api/minimise.md
+++ b/docs/api/minimise.md
@@ -40,27 +40,6 @@ In addition to the following, note that the [Optax](https://github.com/deepmind/
                 - init_hessian
                 - update_hessian
 
-??? abstract "`optimistix.AbstractBFGS`"
-
-    ::: optimistix.AbstractBFGS
-        options:
-            members:
-                false
-
-??? abstract "`optimistix.AbstractDFP`"
-
-    ::: optimistix.AbstractDFP
-        options:
-            members:
-                false
-
-??? abstract "`optimistix.AbstractLBFGS`"
-
-    ::: optimistix.AbstractLBFGS
-        options:
-            members:
-                false
-
 ::: optimistix.BFGS
     options:
         members:
--- a/optimistix/_solver/quasi_newton.py
+++ b/optimistix/_solver/quasi_newton.py
@@ -118,6 +118,10 @@ class AbstractQuasiNewton(
     structure and the Hessian update state, while the latter is called to compute an
     update to the approximation of the Hessian or the inverse Hessian.
 
+    Already supported schemes to form inverse Hessian and Hessian approximations are
+    implemented in `optimistix.AbstractBFGS`, `optimistix.AbstractDFP` and
+    `optimistix.AbstractLBFGS`.
+
     Supports the following `options`:
 
     - `autodiff_mode`: whether to use forward- or reverse-mode autodifferentiation to


tests/helpers.py
```diff
@@ -113,123 +113,123 @@ def __init__(
         self.verbose = frozenset()
 
 
-class BFGSDampedNewton(optx.AbstractQuasiNewton):
+class BFGSDampedNewton(optx.BFGS):
```
patrick-kidger: Can we do this without inheriting from a concrete class? :(

tests/test_lbfgs_linear_operator.py
```diff
@@ -280,9 +280,15 @@ def test_warmup_phase_compact(generate_data, extra_history):
...
-
-    _, state_init = update.init(curr_descent, jnp.array(0.0), curr_descent)
+    solver = optx.LBFGS(
+        rtol=1e-3,
+        atol=1e-6,
+        norm=optx.max_norm,
+        use_inverse=False,
+        history_length=history_len + extra_history,
+    )
+    _, state_init = solver.update_init(curr_descent, jnp.array(0.0), curr_descent)
```
patrick-kidger: The wording of this methods sounds a lot like we're updating (verb) the init (noun), rather than the other way around.

patrick-kidger: Maybe something like `hessian_init` and `hessian_update`? No strong feelings though I like including 'hessian'.

patrick-kidger:
As for private imports in tests, I actually think that is okay. (And is maybe the only place where that is acceptable.) Because they're versioned together then there is no risk of breakage.

tests/helpers.py
johannahaffner: So this goes a little beyond this test module! The methods that update the Hessian approximation are now abstract methods for `AbstractQuasiNewton`, and only the concrete classes provide implementations. 

We could create `Abstract{BFGS, DFP, LBFGS}` that inherit from `AbstractGaussNewton` and provide the method implementations; and then inherit from these to provide `optx.BFGS`. If users want to subclass to make their own solvers, then subclassing e.g. `AbstractBFGS` might be more natural, and tie in better with the abstract/final philosophy (even though it is not enforced anymore). 

If we're not taking this route and want to avoid subclassing a concrete class in this module, then we could implement these inlined methods in the `helpers` file, which would create a lot of code duplication here.

tests/test_lbfgs_linear_operator.py
johannahaffner: Sure! That is an easy fix, and the name is indeed better.

johannahaffner:
I think having `Abstract{BFGS, DFP, LBFGS}` versions make sense here! These can then be subclassed by users who want to create custom solvers. 

Now the hierarchy is as follows:

- `AbstractQuasiNewton` has abstract methods `init_hessian` and `update_hessian`, and concrete methods `init`, `step`, etc.
- `Abstract{BFGS, DFP, LBFGS}` provide implementations for `init_hessian` and `update_hessian`
- Concrete `BFGS`, `DFP`, `LBFGS` specify the search and descent and provide an `__init__`. 

Another thing I noticed is that with the inlined methods, this change is actually not breaking anything with respect to the old `AbstractBFGS` - the attributes are now the same, we only have the `HessianUpdateState` in the (private) solver state now.

I added stubs to the documentation, but when I try to actually add the abstract solvers to the documentation I get the following error: 

```bash
jinja2.exceptions.TemplateNotFound: Tried and failed to find `jax._src.api.ShapeDtypeStruct` in the public API. Commons reasons for this error are (1) if it is from outside this package, then this object is not listed (under whatever public path it should be displayed as) in `hippogriffe.extra_public_objects`; (2) if it is from inside this package, then it may have been written `::: somelib.Foo:` with a trailing colon, when just `:::somelib.Foo` is correct. This arose whilst pretty-printing `optimistix._solver.quasi_newton.AbstractQuasiNewton.init`. You may ignore the rest of this error message, which comes from jinja.
```

This is when putting `AbstractBFGS` below `AbstractQuasiNewton`, and I'm not sure where this is coming from! We do have these structs as part of the solver state, but `init` should not be rendered at all, and is not rendered unless I try to import the documentation for another abstract solver.

PR #148, opened 2025-07-19
Add miscellaneous functions for tree-clipping and mapped where operations

johannahaffner:
This PR will be small once https://github.com/patrick-kidger/optimistix/pull/146 is in. Here is what it adds

- a `tree_clip` function to the miscellaneous module
- test cases to test this function
- overloads the `tree_where` to allow `pred` to be pytree valued and `false` to be a scalar
- test cases to make sure that we handle the possible combinations of `pred`, `true`, `false` correctly

These are needed for the bound-constrained versions of `BFGS` and `LBFGS`, but otherwise not yet used in our main codebase. Of the existing solvers, `NewtonChord` could use `tree_clip`, but I have not made this change. 

The test cases, especially for clipping (a box projection) can also be useful to check if all kinds of projections do the right thing - in this case, only the expected results will have to be altered.

### DIFF
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -83,12 +83,34 @@ def tree_full_like(struct: PyTree, fill_value: ArrayLike, allow_static: bool = F
     return jtu.tree_map(fn, struct)
 
 
+@overload
 def tree_where(
     pred: Bool[ArrayLike, ""], true: PyTree[ArrayLike], false: PyTree[ArrayLike]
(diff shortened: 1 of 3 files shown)


johannahaffner:
Changes: now `pred` can be any tree-prefix of `true` and `false`, while we continue to require the latter to have identical pytree structures. 

This means that we no longer need to overload, and consequently we do not.

Issue #144, opened 2025-06-28
Running mutiple optimizations at once


matillda123:
Hey,

i was wondering how one would use optimistix to run multiple optimizations from different initial guesses. Probably one would use jax.vmap in an interactive solve, right?
Alternatively I was thinking that one could just minimize the sum of the errors from each individual optimization. But maybe this is an issue for higher order methods, since one has a lot of independent variables in this case.

johannahaffner:
You can vmap all Optimistix solves (minimisation, least-squares, root-finds...), this works out of the box without modifications. In an interactive solve like the one [here](https://docs.kidger.site/optimistix/examples/interactive/), you can likewise vmap over the first argument. 

By the way, I remember you were curious about L-BFGS - we now have something you could try in https://github.com/patrick-kidger/optimistix/pull/135.

matillda123:
Then I am doing something wrong? Because it doesnt seem to work with least-squares solvers because of a TypeError from jax.

I tried:

```
import jax.numpy as jnp
import optimistix

import numpy as np


def test_func(x, args):
    return jnp.sum(x**2)

y0 = jnp.array(np.random.uniform(-100, 100, size=(10,10,10)))

solver = optimistix.GaussNewton(rtol=1e-12, atol=1e-12)
solution = jax.vmap(optimistix.minimise, in_axes=(None ,None, 0))(test_func, solver, y0)
```

And I also tried it using:

```
def run_minimise(y0):
    solver = optimistix.GaussNewton(rtol=1e-12, atol=1e-12)
    solution = optimistix.minimise(test_func, solver, y0)
    return solution

solution = jax.vmap(run_minimise, in_axes=0)(y0)
```


> TypeError: Output from batched function { lambda a:f32[10,10]; b:f32[10,10]. let
    c:f32[10,10] = mul b a
    d:f32[] = reduce_sum[axes=(0, 1)] c
  in (d,) } with type <class 'jax._src.core.Jaxpr'> is not a valid JAX type


Similar issues arise when I try this with an interactive solve, where i vmap over solver.init() and solver.step().

I remember that (e.g. when using jax.lax.scan()) one can solve this by using equinox.partition() but I dont see how a user could apply this here.


Ah yes. Ill definitely have a look at LBFGS :)

johannahaffner:
I'm at the lake, quick fix: use Equinox, specifically `eqx.filter_vmap`. This will work, I use it all the time.

[...]

PR #150, opened 2025-07-19
Draft implementation of BFGS-B, L-BFGS-B

johannahaffner:
**Bounded versions of BFGS and L-BFGS**

Adds a draft implementation of BFGS-B and L-BFGS-B, bounded variants of their respective solvers that use a projected active set methods to solve a minimisation problem subject to bounds. The descent method is baptised `CauchyNewtonDescent` for now, and it works as follows:

- At each step, compute the gradient and project its negative (descent directions) onto the faces of the hypercube defined by the bounds
- following the piece-wise linear path defined by the projected gradient, find the first local minimiser of the quadratic approximation to the target function along this path. This point is often termed the "generalised Cauchy point" and we follow this notation here and call it `cauchy_point`. This point is usually found on the surface of the hypercube, but it may also lie in its interior.
- Use the `cauchy_point` to guess which bound constraints are "active": which elements of `y` are at their bounds, and which elements of `y` are not currently blocked / constrained by their bounds? Use this information to decide which bounds to enforce in the next step.
- Take a Newton step over the KKT operator, in which all active bound constraints are imposed as equalities and all other variables are left unconstrained.
- Truncate the Newton step if needed to ensure that we do not leave the bounds (may be necessary if some of the unconstrained elements of `y` are outside of their bounds in the linear solution). This is done such that the overall direction is preserved (i.e. no box-clipping!).
- Return the computed step

We use the same backtracking line search (`BacktrackingArmijo`) that is also used in the unconstrained BFGS-family solvers, so the value returned by the descent specifies a direction along which we may then backtrack. This improves convergence, and since the feasible set is convex in the case of box constraints, there is no risk of leaving it during backtracking.

---

On **naming this method**: this method is introduced under various names in the literature. It is frequently called the "gradient projection method" (e.g. [here](https://epubs.siam.org/doi/epdf/10.1137/0916069), Nocedal & Wright, Numerical Optimisation, Chapter 16.7). Conn, Gould and Toint call it the "Projected Gradient Path" method, and also call the piecewise linear path a "Cauchy arc" (Chapter 12.1.3 in "Trust Region Methods").  They all agree in calling the first minimiser along this path the (generalised) "Cauchy point" $x^c$. 

However, these days projected gradient is more readily understood to refer to first-order methods that do not take a Newton step (and do not perform a linear solve), e.g. [here](https://www.epfl.ch/labs/anchp/wp-content/uploads/2018/05/part6-1.pdf), and the term is often used to describe [proximal gradient methods](https://www.stat.cmu.edu/~ryantibs/convexopt/lectures/prox-grad.pdf), as implemented e.g. in optax under the name ["Projections"](https://optax.readthedocs.io/en/latest/api/projections.html#projections). 

I think it would be useful to at least make `Newton` a part of the name to make it clear that we are expecting to perform a linear solve; and adding `Cauchy` is an attempt to be a bit more descriptive here. `ProjectedNewton` also sounds like we first take a Newton step and then project it.

---

Adding this required a few monkey patches - we have not yet settled on an API for bounded and constrained problems and this implementation is therefore made available as a draft PR only, and uses monkey patches in `AbstractQuasiNewton` to pass the bounds through to the solver. This solver (on this branch only) now accepts `options={"bounds": tuple[Y, Y]}`, where the bounds must be specified as a tuple of pytrees of the same shape as `y`. 

---

The draft implementation is tested on simple cases, and I will add harder test cases and benchmarks later today. It is not yet ready for review. It will have a (considerably) smaller diff if https://github.com/patrick-kidger/optimistix/pull/146, https://github.com/patrick-kidger/optimistix/pull/148, https://github.com/patrick-kidger/optimistix/pull/149 are merged into `dev`.


optimistix/_solver/cauchy_point.py
```diff
@@ -0,0 +1,267 @@
...
+            terminate=terminate,
+        )
+
+    init_state = _CauchyState(intercepts, jnp.array(0.0), y, jnp.array(False))
+    final_state = eqxi.while_loop(cond_fun, body_fun, init_state, kind="lax")
+
+    return final_state.point
+
+
+def cauchy_point(y: Y, f_info: FunctionInfo.EvalGradHessian) -> Y:
```
johannahaffner, line 127: @cjchristopher here we identify the point on the surface of the hypercube that is the first minimiser along the piecewise-linear path of the projected gradient. 

To visualise this, imagine you have some vector pointing out of a cube, and once it passes through a surface you "cut" and allow no further displacement beyond this surface. Then you continue moving along the surface until you meet another surface, and repeat. 

This is different from methods otherwise known as "projected gradient" which project the full gradient step onto the hypercube. In this case you have only two vectors - one is your full gradient, and the other is the result of the clipping operation performed during the box projection. 
For this "projected gradient" method - the name is really not helpful here - we instead get a sequence of vectors that are connected tip-to-tail but do not share directions. This is because we cut the elements in order, instead of all-at-once, and evaluate a quadratic approximation of the target function along this path.

I can draw this if it would be helpful :)

optimistix/_misc.py
```diff
@@ -83,12 +84,127 @@ def tree_full_like(struct: PyTree, fill_value: ArrayLike, allow_static: bool = F
...
+    """
+    return jtu.tree_map(lambda x, l, u: jnp.clip(x, min=l, max=u), tree, lower, upper)
+
+
+def tree_min(tree: PyTree[ArrayLike]) -> Scalar:
+    values, _ = jfu.ravel_pytree(tree)
+    return jnp.min(values)
+
+
+def feasible_step_length(
```
johannahaffner, line 133: Truncate to the maximum feasible (scalar) step length, or obtain a per-leaf maximum feasible step length. Could also be conceptualised as a very simple line search, but I use it more as a small utility. 

Unlike clipping, this does not alter direction - it identifies how far we can move along a proposed path before hitting any bounds, and if that is $0.23 \cdot \Delta x$ for a step $\Delta x$, then this function will tell us that we can move along $\Delta x$ for $0.23$ of its length.

johannahaffner:
> @johannahaffner Seems more sensible to continue discussion here (re #144 )

Indeed! Thanks for the initiative! 

> If you would like to provide some notes there is a (slight) chance I might get time before you do.

Very happy to, would you like an outline of what I think the required steps would be?

> I notice that optimistix currently lacks the various linesearch options that jaxopt does

A Zoom line search is on its way! Here: https://github.com/patrick-kidger/optimistix/pull/143

For the projected steps - for (L-)BFGS-B this refers to finding the Cauchy point, I added a comment pointing to my implementation. I believe it is functionally the same as the JAXopt version [here](https://github.com/google/jaxopt/blob/b12c2f3ddaa3ef71fac42d7e419adccb9dcc49a1/jaxopt/_src/lbfgsb.py#L79), but I have not followed this in every nitty-gritty detail.

The maximum feasible step length is indeed called a line search by some authors, but I just made it a utility function. I pointed out its location with another comment.

> Separately, Is the projected gradient descent in #119 currently usable? Perhaps not the ideal choice for my problem, but I'd be interested to see how it does. 

This branch is super old and somewhat stale, but the wrapper is very simple and I'm happy to pull it out of its misery in that old draft branch and put it in a fresh one for you to try :)

Uripeer3:
@johannahaffner hey, I've been looking around the past few days for a JAX-based BFGS-B optimizer and wandered off here.
I was wondering why this lead closed, and if maybe I can help ^_^

johannahaffner:
It is closed because I'm currently very very busy with non-Github work :) If you want to help, I'm very happy to hand this over! Bounded BFGS is a small project, bounded L-BFGS is more complex, since it will need approximations of both the inverse Hessian and the Hessian and these will need to be pipelined in a new solver, rather than inheriting from `AbstractQuasiNewton`, which currently supports using _either_ the Hessian or the inverse Hessian, but not both (and should `AbstractQuasiNewton` should keep enforcing this separation, which is much cleaner and more memory-efficient for all use cases outside of `L-BFGS`). 

### Bounded BFGS:

This needs a different descent that can then be used with `QuasiNewton` solvers that provide a Hessian approximation. The new descent should do the following:

1. At every trial iterate, project the gradient onto the bounds using the `cauchy_point` function already in `main`: https://github.com/patrick-kidger/optimistix/blob/8dc7f30bffcf194445a03b2ecb71b935f56f8d22/optimistix/_solver/cauchy_point.py#L135

I would say that this function is the trickiest bit of the solver, and it has been very thoroughly tested already. 

2. Use the Cauchy point to fix certain elements of `y` to their bounds, and optimise over the others, by defining the appropriate KKT system. The following works:
```python
def cauchy_newton_step(y, f_info, linear_solver):
    cauchy = cauchy_point(y, f_info)
    lower, upper = f_info.bounds  # pyright: ignore (handled by cauchy_point)

    active_lower = jtu.tree_map(lambda a, b: a <= b, cauchy, lower)
    active_upper = jtu.tree_map(lambda a, b: a >= b, cauchy, upper)
    active_bound_values = tree_where(active_lower, lower, tree_full_like(lower, 0.0))
    active_bound_values = tree_where(active_upper, upper, active_bound_values)
    active = jtu.tree_map(lambda a, b: a | b, active_lower, active_upper)

    bound_constraint = lx.DiagonalLinearOperator(active)
    dual_constraint = lx.DiagonalLinearOperator(jtu.tree_map(jnp.invert, active))

    def make_kkt_function(hessian, bound_constraint, dual_constraint):
        def kkt_function(inputs):
            y, dual = inputs
            y_pred = (hessian.mv(y) ** ω - bound_constraint.transpose().mv(dual) ** ω).ω
            dual_pred = (bound_constraint.mv(y) ** ω + dual_constraint.mv(dual) ** ω).ω
            return y_pred, dual_pred

        return kkt_function

    kkt_function = make_kkt_function(f_info.hessian, bound_constraint, dual_constraint)
    in_struct = jax.eval_shape(lambda: (y, y))
    kkt_operator = lx.FunctionLinearOperator(kkt_function, in_struct)

    kkt_function = make_kkt_function(f_info.hessian, bound_constraint, dual_constraint)
    in_struct = jax.eval_shape(lambda: (y, y))
    kkt_operator = lx.FunctionLinearOperator(kkt_function, in_struct)

    # Exact dual values: grad(f) = lambda * grad(c), where c is the constraint function.
    # Since we only have bound constraints, grad(c) is one where the constraint is
    # active and zero where it is not.
    dual = tree_where(active, f_info.grad, 0.0)
    upper_rhs = (bound_constraint.mv(dual) ** ω - f_info.grad**ω).ω
    lower_rhs = (active_bound_values**ω - bound_constraint.mv(y) ** ω).ω
    vector = upper_rhs, lower_rhs

    out = lx.linear_solve(kkt_operator, vector, linear_solver)
    cauchy_newton, _ = out.value
    result = RESULTS.promote(out.result)

    max_step_size = feasible_step_length(y, cauchy_newton, lower, upper)
    cauchy_newton = (max_step_size * cauchy_newton**ω).ω

    return cauchy_newton, result
```

And the corresponding descent is something like 

```python
class CauchyNewtonDescent(
    AbstractDescent[
        Y,
        Iterate.Primal,  # TODO: currently unused
        FunctionInfo.EvalGradHessian,
        _CauchyNewtonDescentState,
    ],
):
    """Computes the Cauchy point to make an educated guess of the set of active bound
    constraints. Then, a Newton step is computed using the KKT system, where the active
    bound constraints are enforced as equalities.

    This means that we are taking a traditional Newton step - towards the bottom of a
    quadratic bowl - in the unconstrained elements of `y`, and a projected gradient step
    in the constrained elements of `y`.
    """
    linear_solver: lx.AbstractLinearSolver = lx.AutoLinearSolver(well_posed=False)

    def init(  # pyright: ignore  TODO
        self, y: Y, f_info_struct: FunctionInfo.EvalGradHessian
    ) -> _CauchyNewtonDescentState:
        del f_info_struct
        return _CauchyNewtonDescentState(cauchy_newton=y, result=RESULTS.successful)

    def query(  # pyright: ignore  TODO
        self,
        y: Y,
        f_info: FunctionInfo.EvalGradHessian,
        state: _CauchyNewtonDescentState,
    ) -> _CauchyNewtonDescentState:
        del state
        cauchy_newton, result = cauchy_newton_step(y, f_info, self.linear_solver)
        return _CauchyNewtonDescentState(cauchy_newton=cauchy_newton, result=result)

    def step(  # pyright: ignore TODO
        self, step_size: Scalar, state: _CauchyNewtonDescentState
    ) -> tuple[Y, RESULTS]:
        # TODO(jhaffner): double check signs!
        return (step_size * state.cauchy_newton**ω).ω, state.result
```

3. Combine the new descent with e.g. BFGS to make a new public solver, e.g. as is done here: https://github.com/patrick-kidger/optimistix/blob/8dc7f30bffcf194445a03b2ecb71b935f56f8d22/optimistix/_solver/quasi_newton.py#L418

This does not require any changes to the Hessian approximation. An appropriate search to combine this with **must** guarantee that the step length is at most `1.0` to avoid leaving the bounded region. A simple choice would be `LearningRate(1.0)`, also used in `GaussNewton`, but `BacktrackingArmijo` also works and can be more robust. The new descent should clearly document that it should not be used with a search such as `ClassicalTrustRegion` that may produce step sizes that are larger than one. 

4. Find an elegant / useful way to communicate to the solver/descent what the bounds are. Currently these are accepted as `options`, e.g. here: https://docs.kidger.site/optimistix/api/root_find/#optimistix.Newton, but we were thinking about ways to make a judicious API break here to support more general solvers. Unsolved API questions that are much more relevant for solvers that are substantially more complicated than this one are actually what stopped the release of this one. So this is a question that needs settling before this could be merged. 

_If you want to play around with it for your own use case/code, there are plenty of options - easiest would be to make the descent accept bounds in the constructor so they are available whenever you need them (which is nowhere else!) and then you won't have to worry about handling function information objects or indeed knowing any part of the Optimistix code base outside of the new descent, although you would have to construct a new solver each time you want to change the bounds._

5. Adding tests - there are lots of test cases, smokey and otherwise, defined in https://github.com/patrick-kidger/optimistix/blob/main/tests/helpers.py already. 

**Minor doc updates**: 

- `cauchy_point`, due to the internal `while_loop`, is currently not compatible with `optx.RecursiveCheckpointAdjoint`. This should be noted in the docs for this function, and it should be noted in the docs for any public solver that builds on it. Using `RecursiveCheckpointAdjoint` for nonlinear optimisation is pretty niche, and making it compatible would be a pain (it would entail forwarding the adjoint method somehow), likely without any numerical gain. 

**Subtler points**

In the `cauchy_newton_step` function above, I compute the values of the multipliers on the bound constraints from scratch at every iteration, without maintaining these as separate iterates - and without updating them using their computed steps. When I wrote this, I convinced myself that this was permissible with pen and paper, but that was a while ago and I should go back and look at it again.

Issue #152, opened 2025-07-26
Feature Request: SQUAREM Accelerated Fixed Point Methods


AndrewCSQ:
Thanks for the fantastic library! I have a memory sensitive application that requires me to prefer iterative methods for my fixed point solve rather than Jacobian based methods. [SQUAREM](https://www.jstatsoft.org/article/view/v092i07/1338) is a fixed point acceleration method that effectively approximates the Jacobian sans memory requirements.

Before the advent of `optimistix` I effectively implemented SQUAREM with an `optimistix`-like interface, sans the actually robust software engineering (I am an economist with minimal formal software engineering training). I'm looking to upstream SQUAREM within `optimistix` as a replacement.

## Description of Algorithm

SQUAREM modifies the naive iteration update step with the following. For a given fixed point equation $F(\theta)$, run the function $F$ twice. Then apply a steplength $\alpha$ based on these two updates, and run the function a final time. The pseudocode references the original 2008 SQUAREM paper, which unfortunately is not open access.

<img width="920" height="706" alt="Image" src="https://github.com/user-attachments/assets/4916b964-a734-464d-9ede-800462e4341b" />

The authors suggest three possibilities for the steplength $\alpha$, S1 to S3. S3 is recommended.

<img width="250" height="127" alt="Image" src="https://github.com/user-attachments/assets/f435928a-4c16-42d7-af08-41f1fcb473f8" />

<img width="250" height="127" alt="Image" src="https://github.com/user-attachments/assets/8f043a1a-debf-4013-b7d7-23a689de2831" />

<img width="250" height="127" alt="Image" src="https://github.com/user-attachments/assets/e9e80e4f-dac5-4ba9-8112-5aa17b22e5cb" />

## Proposed Implementation

I've implemented this by modifying the `_solver/fixed_point.py` file to:
1. Include S1-S3 within the class
2. Implement the pseudocode above for the step function

This is my first foray into playing with the internals of a "proper" library and
1. I really have no clue what the $\omega$ in the code are supposed to do (I understand from the `equinox` library that it has something to do with tree-mappings, and that has something to do with the graph theory underpinning JAX?)
2. I wasn't sure how to initialise the stepsize as a parameter (which it should be), so like any amateur code conjurer I've hard-coded the update function to use scheme 3.

```
class Squarem(
    AbstractFixedPointSolver[Y, Aux, _FixedPointState], strict=True
):
    """
    Uses SQUAREM acceleration
    """
[...]

PR #153, opened 2025-07-28
SQUAREM Accelerated Fixed Point Method

AndrewCSQ:
Relevant [issue](https://github.com/patrick-kidger/optimistix/issues/152)

Some memory-sensitive applications prefer avoiding Jacobian-based methods for solving fixed-point equations. SQUAREM (Varadhan and Roland 2008; see R [implementation](https://www.jstatsoft.org/article/view/v092i07/1338)) is an acceleration scheme that approximates the Jacobian using iterates of the fixed point equation.

## Description of Algorithm

The pseudocode references the original 2008 SQUAREM paper, which unfortunately is not open access.

Image

## Solver Settings

The authors suggest three possibilities for the steplength $\alpha$, S1 to S3. S3 is recommended.

Image

Image

Image

## Scope of PR

I took inspiration from the existing `fixed_point.py` and the [Anderson Acceleration](https://github.com/patrick-kidger/optimistix/pull/145) PR.

- [x] Created new file `squarem.py` under the `_solver` folder.
- [x] Imported the new `Squarem` class under the `__init__.py` for both `_solver` and the main `optimistix` `__init__.py`
- [x] Added `Squarem` as a method for tests under `tests/test_fixed_point.py`
- [ ] Make the steplength scheme a configurable parameter. I wasn't quite sure how to modify the `__init__` function for the `AbstractFixedPointSolver` class. For now, S3 is hardcoded
- [ ] Proper typing for the steplength schemes. I trial-and-errored `Y` until the pre-commit checks passed.
- [ ] Any documentation changes?


optimistix/__init__.py
```diff
@@ -75,6 +75,7 @@
     NonlinearCGDescent as NonlinearCGDescent,
     OptaxMinimiser as OptaxMinimiser,
     polak_ribiere as polak_ribiere,
+    Squarem as Squarem,
```
johannahaffner: Do you think we should capitalise this? Or go with `SquarEM` to emphasise the connection to `EM` methods?

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+from .._fixed_point import AbstractFixedPointSolver
+from .._misc import max_norm
+from .._solution import RESULTS
+from .._solver.fixed_point import _FixedPointState
+
+
+# optimistix compliant implementation of SQUAREM accelerated fixed point
+class Squarem(AbstractFixedPointSolver[Y, Aux, _FixedPointState], strict=True):
+    r"""
+    SQUAREM acceleration for fixed point equations per Varadhan and Roland (2008).
```
johannahaffner: Nit: no newline here.

johannahaffner: I think it would be nice to include the original reference here. You can do this following this example: https://github.com/patrick-kidger/diffrax/blob/619445c15d802775d08adfb57e4c3a7181794a04/diffrax/_adjoint.py#L196

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+    def s3(r: Y, v: Y):
+        r"""Scheme 3 of SQUAREM algorithm steplength
+
+        Equation (9) of Varadhan and Roland (2008)
+        for the computation of $\alpha = \frac{\left\|r^Tr\right\|{\left\|r^Tv\right\|$.
+        """
+        return (
+            -1
+            * (
+                ω(r).call(lambda x: jnp.linalg.norm(x, ord=2))
```
johannahaffner: Breaking this up would make it easier to read: maybe take the `ω(r).call(lambda x: jnp.linalg.norm(x, ord=2))` out of the return statement and give them a descriptive name.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+                / ω(v).call(lambda x: jnp.linalg.norm(x, ord=2))
+            ).ω
+        )
+
+    rtol: float
```
johannahaffner, line 75: Nit: move attribute declarations above method definitions, directly below the docstring.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+        self,
+        fn: Fn[Y, Y, Aux],
+        y: Y,
+        args: PyTree,
+        options: dict[str, Any],
+        state: _FixedPointState,
+        tags: frozenset[object],
+        # scheme is hard-coded for now
+        # hopefully the Optimistix maintainers will show how to put this into init
+        scheme=s3,
```
johannahaffner: This should be set during in the `__init__` method for this particular solver. This solver currently does not have one (yet). Take a look at `optx.NonlinearCG` to see how this can be done. I think we can make the schemes available as methods in a similar fashion. 

The alternative would be to make them searches, e.g. like `optx.LearningRate`, but I'm not sure that this would fit with their current signatures.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+    ) -> tuple[Y, _FixedPointState, Aux]:
+        # step twice
+        y1, _ = fn(y, args)
+        y2, _ = fn(y, args)
+        # compute r, v in the paper
+        r = (y1**ω - y**ω).ω
```
johannahaffner, line 131: Since you left a comment about the `ω`: these are little helpers for arithmetic on pytrees :)

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,144 @@
...
+        args: PyTree,
+        options: dict[str, Any],
+        state: _FixedPointState,
+        tags: frozenset[object],
+        # scheme is hard-coded for now
+        # hopefully the Optimistix maintainers will show how to put this into init
+        scheme=s3,
+    ) -> tuple[Y, _FixedPointState, Aux]:
+        # step twice
+        y1, _ = fn(y, args)
```
johannahaffner, line 128: `y1` and `y2` are going to take identical values here, I think that is not intended?

johannahaffner (commented):
I've left a first round of comments :)

optimistix/__init__.py
AndrewCSQ: Am okay to go with SquarEM

optimistix/_solver/squarem.py
AndrewCSQ: Amended.

AndrewCSQ: Had a go at including the original reference. See if okay?

optimistix/_solver/squarem.py
AndrewCSQ: Have tried to do so by defining the L2 norm of r and v separately. See if this is okay?

optimistix/_solver/squarem.py
AndrewCSQ: amended please

optimistix/_solver/squarem.py
AndrewCSQ: Thanks for the reference to `optx.NonLinearCG`. Have taken a stab at an `__init__` method accordingly, see if okay? Agreed we should view them as methods rather than signatures.

optimistix/_solver/squarem.py
AndrewCSQ: Ah sorry this was a bout of carelessness. Have corrected (so the code now actually accelerates)

AndrewCSQ:
Thanks for the initial round of comments, have amended accordingly.

AndrewCSQ:
Looking through the tests, I think there's something wrong with the implementation (not converging on the fixed points). Please let me look through the implementation again (and compare against other implementations elsewhere). Will ping when it's ready for review.

AndrewCSQ:
Sorry for yet another change. Reading through the original paper and have incorporated Section 6's proposal to ensure that the size of the steplength is at least 1 (i.e., that steplength < -1). This breaks  quite a few more complex-valued parameter tests, but improves performance for my real-valued parameter searches.

I think we can add documentation that the current SquarEM implementation should only be used for real-valued parameters. I don't have the math chops to extend the paper to the complex-valued case, and a cursory search didn't turn up anything either.

johannahaffner:
Alright, I've completed the rebase and we now have a clean history with the `SquarEM` implementation in a single commit attributed to you :) 

I get the same test failures you reported earlier, will dive into that now.

AndrewCSQ:
Let me know what I can do to push this PR along (hoping for this to make it into the next `optimistix` release so I don't have to maintain the SquarEM implementation out-of-tree for my own library). My use-case only involves real-valued vectors, but I'm happy to help work through the complex-valued arbitrary array case.

johannahaffner:
I'm sorry that I'm taking so long on this one! We can get this one in before the next release :) 

We will probably wait until JAX 0.7.2 is out, since 0.7.0 and 0.7.1 introduced some performance regressions for which 0.7.2 will contain fixes. Would that work for you?

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+from jaxtyping import Array, Bool, PyTree, Scalar
+
+from .._custom_types import Aux, Fn, Y
+from .._fixed_point import AbstractFixedPointSolver
+from .._misc import max_norm
+from .._solution import RESULTS
+from .._solver.fixed_point import _FixedPointState
+
+
+# although Varadhan and Roland (2008) provide 3 different schemes
```
johannahaffner, line 17: I think we can generalise this to the arbitrarily-shaped case: the `tree_dot` function from the `optx._misc` module does exactly that. 

Every arbitrarily dimensioned PyTree of arrays can always be flattened to the 1D case, and hence such operations are always possible. 

The schemes could be implemented as

```python
from .._misc import tree_dot

def s1(r: Y, y: Y) -> Scalar: 
    return tree_dot(r, v) / tree_dot(v, v)
```

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+    l2_r = ω(r).call(lambda x: jnp.linalg.norm(x, ord=None))
+    l2_v = ω(v).call(lambda x: jnp.linalg.norm(x, ord=None))
+    return (-1.0 * l2_r / l2_v).ω
+
+
+# optimistix compliant implementation of SQUAREM accelerated fixed point
+class SquarEM(AbstractFixedPointSolver[Y, Aux, _FixedPointState]):
+    r"""SQUAREM acceleration for fixed point equations per Varadhan and Roland (2008).
+    As we are using it purely for fixed point acceleration, we do not implement
+    the option of passing in a likelihood function (useful when SQUAREM is used for EM
```
johannahaffner, line 56: I'm curious about the likelihood function - they write that although it is not needed, it is required to guarantee global convergence. Looking at the pseudocode in from the software package paper (Du and Varadhan) it seems like providing it would be quite simple: the users could pass a custom callable that takes `y` as its sole input and generates a scalar value, and this can then be used to select between `y_2` and `y_prime`. 

Am I missing something here? This could look something like 

```python
class SquarEM(...):
    likelihood: Callable[[Y], Scalar] | None = None
    ...
    def step(...):
        ...
        if self.likelihood is not None:
             # Select y2 if self.likelihood(y2) > self.likelihood(y_prime) - eta
        else:
             ...
```

Now, I'm not sure what exactly `eta` is, I'm assuming that it is some kind of tolerance (I couldn't find it in the paper, unfortunately).

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+
+        - `rtol`: Relative tolerance for terminating solve.
+        - `atol`: Absolute tolerance for terminating solve.
+        - `norm`: The norm used to determine the difference between two iterates in the
+            convergence criteria. Should be any function `PyTree -> Scalar`. Optimistix
+            includes three built-in norms: [`optimistix.max_norm`][],
+            [`optimistix.rms_norm`][], and [`optimistix.two_norm`][].
+        - `method`: The function which computes the steplength `alpha`.
+            Only Scheme 3 using Frobenius norms is implemented as the
+            other schemes do not generalise to parameters with many
```
johannahaffner, line 97: If we implement the schemes as suggested above, then these lines will be outdated.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+        y1, _ = fn(y, args)
+        y2, _ = fn(y1, args)
+        # compute r, v in the paper
+        r = (y1**ω - y**ω).ω
+        v = (y2**ω - y1**ω - r**ω).ω
+        # compute the alpha based on scheme chosen at initialisation
+        with jax.numpy_dtype_promotion("standard"):
+            alpha = self.scheme(r, v)
+            # per Varadhan and Roland (2008), set the stepsize to be at most -1
+            alpha = ω(alpha).call(lambda x: jnp.minimum(x, -1.0))
```
johannahaffner, line 137: `alpha` should be a scalar, shouldn't it? Hence the `ω` should not be necessary. Let's see if pyright still complains after we've replaced the schemes with the `tree_dot` implementations. 

If it does, maybe a `jnp.asarray` is all it will take.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+
+SquarEM.__init__.__doc__ = """**Arguments:**
+
+- `rtol`: Relative tolerance for terminating the solve.
+- `atol`: Absolute tolerance for terminating the solve.
+- `norm`: The norm used to determine the difference between two iterates in the 
+    convergence criteria. Should be any function `PyTree -> Scalar`. Optimistix
+    includes three built-in norms: [`optimistix.max_norm`][],
+    [`optimistix.rms_norm`][], and [`optimistix.two_norm`][].
+- `scheme`: Scheme used to compute the steplength $alpha$.
```
johannahaffner, line 180: Will also need updating once we have the other schemes back in :)

optimistix/_solver/squarem.py
AndrewCSQ: I tried this out (for `s2`) and got a number of test-case failures. I think it has to do with the "more advanced" cases where a pytree is passed in rather than a standard parameter vector

```
=================================== FAILURES ===================================
__________________ test_fixed_point[_cos-init0-args0-solver1] __________________

solver = SquarEM(rtol=1e-06, atol=1e-06)
_fn = <function _cos at 0x7f9d90381440>
init = ({'a': Array([[[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, 0.5, 0.5]],

       [[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, ...    [[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, 0.5, 0.5]]], dtype=float64)}, Array([0.5, 0.5, 0.5, 0.5], dtype=float64))
args = Array(1., dtype=float64, weak_type=True)

    @pytest.mark.parametrize("solver", _fp_solvers)
    @pytest.mark.parametrize("_fn, init, args", fixed_point_fn_init_args)
    def test_fixed_point(solver, _fn, init, args):
        atol = rtol = 1e-4
        has_aux = random.choice([True, False])
        if has_aux:
            fn = lambda x, args: (_fn(x, args), smoke_aux)
        else:
            fn = _fn
>       optx_fp = optx.fixed_point(
            fn, solver, init, has_aux=has_aux, args=args, max_steps=10_000, throw=False
        ).value

tests/test_fixed_point.py:41: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
optimistix/_fixed_point.py:141: in fixed_point
    return iterative_solve(
optimistix/_iterate.py:344: in iterative_solve
    ) = adjoint.apply(_iterate, rewrite_fn, inputs, tags)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_adjoint.py:133: in apply
    return implicit_jvp(primal_fn, rewrite_fn, inputs, tags, self.linear_solver)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_ad.py:60: in implicit_jvp
    root, residual = _implicit_impl(fn_primal, fn_rewrite, inputs, tags, linear_solver)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_ad.py:67: in _implicit_impl
    return jtu.tree_map(jnp.asarray, fn_primal(inputs))
                                     ^^^^^^^^^^^^^^^^^
optimistix/_iterate.py:240: in _iterate
    final_carry = while_loop(cond_fun, body_fun, init_carry, max_steps=max_steps)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/equinox/internal/_loop/loop.py:103: in while_loop
    _, _, _, final_val = lax.while_loop(cond_fun_, body_fun_, init_val_)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/equinox/internal/_loop/common.py:471: in new_body_fun
    buffer_val2 = body_fun(buffer_val)
                  ^^^^^^^^^^^^^^^^^^^^
optimistix/_iterate.py:230: in body_fun
    new_y, new_state, aux = solver.step(fn, y, args, options, state, tags)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_solver/squarem.py:140: in step
    y_prime = (y**ω - 2.0 * alpha * r**ω + alpha**2 * v**ω).ω
                      ^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = ω(Traced<float64[]>with<DynamicJaxprTrace>)
other = ω(({'a': Traced<float64[3,2,4]>with<DynamicJaxprTrace>}, Traced<float64[4]>with<DynamicJaxprTrace>))

    def fn(self, other):
        if isinstance(other, ω):
            if jtu.tree_structure(self.ω) != jtu.tree_structure(other.ω):
>               raise ValueError("PyTree structures must match.")
E               ValueError: PyTree structures must match.
E               --------------------
E               For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

optimistix/_solver/squarem.py
AndrewCSQ: Similar to trying to use `tree_dot`, when I tried this I got some typing errors. I coded this as

```
alpha = self.scheme(r, v)
alpha = jnp.minimum(jnp.asarray(alpha).flatten(), jnp.asarray([-1.]))
```

```
=================================== FAILURES ===================================
__________________ test_fixed_point[_cos-init0-args0-solver1] __________________

solver = SquarEM(rtol=1e-06, atol=1e-06)
_fn = <function _cos at 0x7f0d303994e0>
init = ({'a': Array([[[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, 0.5, 0.5]],

       [[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, ...    [[0.5, 0.5, 0.5, 0.5],
        [0.5, 0.5, 0.5, 0.5]]], dtype=float64)}, Array([0.5, 0.5, 0.5, 0.5], dtype=float64))
args = Array(1., dtype=float64, weak_type=True)

    @pytest.mark.parametrize("solver", _fp_solvers)
    @pytest.mark.parametrize("_fn, init, args", fixed_point_fn_init_args)
    def test_fixed_point(solver, _fn, init, args):
        atol = rtol = 1e-4
        has_aux = random.choice([True, False])
        if has_aux:
            fn = lambda x, args: (_fn(x, args), smoke_aux)
        else:
            fn = _fn
>       optx_fp = optx.fixed_point(
            fn, solver, init, has_aux=has_aux, args=args, max_steps=10_000, throw=False
        ).value

tests/test_fixed_point.py:41: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
optimistix/_fixed_point.py:141: in fixed_point
    return iterative_solve(
optimistix/_iterate.py:344: in iterative_solve
    ) = adjoint.apply(_iterate, rewrite_fn, inputs, tags)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_adjoint.py:133: in apply
    return implicit_jvp(primal_fn, rewrite_fn, inputs, tags, self.linear_solver)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_ad.py:60: in implicit_jvp
    root, residual = _implicit_impl(fn_primal, fn_rewrite, inputs, tags, linear_solver)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_ad.py:67: in _implicit_impl
    return jtu.tree_map(jnp.asarray, fn_primal(inputs))
                                     ^^^^^^^^^^^^^^^^^
optimistix/_iterate.py:240: in _iterate
    final_carry = while_loop(cond_fun, body_fun, init_carry, max_steps=max_steps)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/equinox/internal/_loop/loop.py:103: in while_loop
    _, _, _, final_val = lax.while_loop(cond_fun_, body_fun_, init_val_)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/equinox/internal/_loop/common.py:471: in new_body_fun
    buffer_val2 = body_fun(buffer_val)
                  ^^^^^^^^^^^^^^^^^^^^
optimistix/_iterate.py:230: in body_fun
    new_y, new_state, aux = solver.step(fn, y, args, options, state, tags)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
optimistix/_solver/squarem.py:137: in step
    alpha = jnp.minimum(jnp.asarray(alpha).flatten(), jnp.asarray([-1.]))
                        ^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/jax/_src/numpy/array.py:383: in asarray
    return array(a, dtype=dtype, copy=bool(copy), order=order, device=device)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/jax/_src/numpy/array.py:251: in array
    arrays_out = [lax.expand_dims(arr, [0]) for arr in arrs]
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../../.local/lib/python3.13/site-packages/jax/_src/numpy/array.py:250: in <genexpr>
    arrs = (array(elt, dtype=dtype, copy=False) for elt in object)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

object = {'a': Traced<float64[]>with<DynamicJaxprTrace>}
dtype = dtype('float64'), copy = False, order = 'K', ndmin = 0

    @export
    def array(object: Any, dtype: DTypeLike | None = None, copy: bool = True,
              order: str | None = "K", ndmin: int = 0,
              *, device: xc.Device | Sharding | None = None) -> Array:
      """Convert an object to a JAX array.
    
      JAX implementation of :func:`numpy.array`.
    
      Args:
        object: an object that is convertible to an array. This includes JAX
          arrays, NumPy arrays, Python scalars, Python collections like lists
          and tuples, objects with an ``__array__`` method, and objects
          supporting the Python buffer protocol.
        dtype: optionally specify the dtype of the output array. If not
          specified it will be inferred from the input.
        copy: specify whether to force a copy of the input. Default: True.
        order: not implemented in JAX
        ndmin: integer specifying the minimum number of dimensions in the
          output array.
        device: optional :class:`~jax.Device` or :class:`~jax.sharding.Sharding`
          to which the created array will be committed.
    
      Returns:
        A JAX array constructed from the input.
    
      See also:
        - :func:`jax.numpy.asarray`: like `array`, but by default only copies
          when necessary.
        - :func:`jax.numpy.from_dlpack`: construct a JAX array from an object
          that implements the dlpack interface.
        - :func:`jax.numpy.frombuffer`: construct a JAX array from an object
          that implements the buffer interface.
    
      Examples:
        Constructing JAX arrays from Python scalars:
    
        >>> jnp.array(True)
        Array(True, dtype=bool)
        >>> jnp.array(42)
        Array(42, dtype=int32, weak_type=True)
        >>> jnp.array(3.5)
        Array(3.5, dtype=float32, weak_type=True)
        >>> jnp.array(1 + 1j)
        Array(1.+1.j, dtype=complex64, weak_type=True)
    
        Constructing JAX arrays from Python collections:
    
        >>> jnp.array([1, 2, 3])  # list of ints -> 1D array
        Array([1, 2, 3], dtype=int32)
        >>> jnp.array([(1, 2, 3), (4, 5, 6)])  # list of tuples of ints -> 2D array
        Array([[1, 2, 3],
               [4, 5, 6]], dtype=int32)
        >>> jnp.array(range(5))
        Array([0, 1, 2, 3, 4], dtype=int32)
    
        Constructing JAX arrays from NumPy arrays:
    
        >>> jnp.array(np.linspace(0, 2, 5))
        Array([0. , 0.5, 1. , 1.5, 2. ], dtype=float32)
    
        Constructing a JAX array via the Python buffer interface, using Python's
        built-in :mod:`array` module.
    
        >>> from array import array
        >>> pybuffer = array('i', [2, 3, 5, 7])
        >>> jnp.array(pybuffer)
        Array([2, 3, 5, 7], dtype=int32)
      """
      if order is not None and order != "K":
        raise NotImplementedError("Only implemented for order='K'")
    
      # check if the given dtype is compatible with JAX
      dtypes.check_user_dtype_supported(dtype, "array")
    
      # Here we make a judgment call: we only return a weakly-typed array when the
      # input object itself is weakly typed. That ensures asarray(x) is a no-op
      # whenever x is weak, but avoids introducing weak types with something like
      # array([1, 2, 3])
      weak_type = dtype is None and dtypes.is_weakly_typed(object)
      if device is None and isinstance(object, core.Tracer):
        sharding = object.aval.sharding
        sharding = None if sharding.mesh.empty else sharding
      else:
        sharding = util.canonicalize_device_to_sharding(device)
    
      # Use device_put to avoid a copy for ndarray inputs.
      if (not copy and isinstance(object, np.ndarray) and
          (dtype is None or dtype == object.dtype) and (ndmin <= object.ndim) and
          device is None):
        # Keep the output uncommitted.
        return api.device_put(object)
    
      # String arrays need separate handling because XLA does not support string
      # as a data type.
      if dtypes.is_string_dtype(dtype) or (
          hasattr(object, "dtype") and dtypes.is_string_dtype(object.dtype)
      ):
        return _make_string_array(
            object=object, dtype=dtype, ndmin=ndmin, device=device
        )
    
      # For Python scalar literals, call coerce_to_array to catch any overflow
      # errors. We don't use dtypes.is_python_scalar because we don't want this
      # triggering for traced values. We do this here because it matters whether or
      # not dtype is None. We don't assign the result because we want the raw object
      # to be used for type inference below.
      if isinstance(object, (bool, int, float, complex)):
        _ = dtypes.coerce_to_array(object, dtype)
      elif not isinstance(object, Array):
        # Check if object supports any of the data exchange protocols
        # (except dlpack, see data-apis/array-api#301). If it does,
        # consume the object as jax array and continue (but not return) so
        # that other array() arguments get processed against the input
        # object.
        #
        # Notice that data exchange protocols define dtype in the
        # corresponding data structures and it may not be available as
        # object.dtype. So, we'll resolve the protocols here before
        # evaluating object.dtype.
        if hasattr(object, '__jax_array__'):
          object = object.__jax_array__()
        elif hasattr(object, '__cuda_array_interface__'):
          cai = object.__cuda_array_interface__
          backend = xla_bridge.get_backend("cuda")
          if cuda_plugin_extension is None:
            device_id = None
          else:
            device_id = cuda_plugin_extension.get_device_ordinal(cai["data"][0])
          object = xc._xla.cuda_array_interface_to_buffer(
              cai=cai, gpu_backend=backend, device_id=device_id)
    
      leaves, treedef = tree_util.tree_flatten(object, is_leaf=lambda x: x is None)
      if any(leaf is None for leaf in leaves):
        raise ValueError("None is not a valid value for jnp.array")
      leaves = [
          leaf
          if (leaf_jax_array := getattr(leaf, "__jax_array__", None)) is None
          else leaf_jax_array()
          for leaf in leaves
      ]
      if dtype is None:
        # Use lattice_result_type rather than result_type to avoid canonicalization.
        # Otherwise, weakly-typed inputs would have their dtypes canonicalized.
        try:
          dtype = dtypes._lattice_result_type(*leaves)[0] if leaves else dtypes.float_
        except TypeError:
          # This happens if, e.g. one of the entries is a memoryview object.
          # This is rare, so we only handle it if the normal path fails.
          leaves = [_convert_to_array_if_dtype_fails(leaf) for leaf in leaves]
          dtype = dtypes._lattice_result_type(*leaves)[0]
    
      if not weak_type:
        dtype = dtypes.canonicalize_dtype(dtype, allow_extended_dtype=True)  # type: ignore[assignment]
    
      object = treedef.unflatten(leaves)
      out: ArrayLike
      if all(not isinstance(leaf, Array) for leaf in leaves):
        # TODO(jakevdp): falling back to numpy here fails to overflow for lists
        # containing large integers; see discussion in
        # https://github.com/jax-ml/jax/pull/6047. More correct would be to call
        # coerce_to_array on each leaf, but this may have performance implications.
        out = np.asarray(object, dtype=dtype)
      elif isinstance(object, Array):
        assert object.aval is not None
        out = lax._array_copy(object) if copy else object
      elif isinstance(object, (list, tuple)):
        if object:
          arrs = (array(elt, dtype=dtype, copy=False) for elt in object)
          arrays_out = [lax.expand_dims(arr, [0]) for arr in arrs]
          # lax.concatenate can be slow to compile for wide concatenations, so form a
          # tree of concatenations as a workaround especially for op-by-op mode.
          # (https://github.com/jax-ml/jax/issues/653).
          k = 16
          while len(arrays_out) > k:
            arrays_out = [lax.concatenate(arrays_out[i:i+k], 0)
                          for i in range(0, len(arrays_out), k)]
          out = lax.concatenate(arrays_out, 0)
        else:
          out = np.array([], dtype=dtype)
      elif _supports_buffer_protocol(object):
        object = memoryview(object)
        # TODO(jakevdp): update this once we support NumPy 2.0 semantics for the copy arg.
        out = np.array(object) if copy else np.asarray(object)
      else:
>       raise TypeError(f"Unexpected input type for array: {type(object)}")
E       TypeError: Unexpected input type for array: <class 'dict'>
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

optimistix/_solver/squarem.py
AndrewCSQ: Thanks for this. I can pass you a copy of the original 2008 paper over email if needed, but I think the more productive way to chase this "stabilisation step" down would be to inspect the [source code](https://cran.r-project.org/src/contrib/SQUAREM_2021.1.tar.gz) of the `R` package. Untar, then look in the `R` folder, all the code is in a single `.R` file.

I think the most relevant part of the R code would be the `squarem2` function that deals with contraction mappings, and it seems like the authors implemented a more sophisticated strategy than just setting to -1. Note that I think they've also done a sign-flip so `alpha` is positive.

The comment about min/max step sizes being linked to the sign of the eigenvalues of the Jacobian of the fixed-point mapping could address the complex-valued case - this is lower priority for me to explore but happy to take guidance.

I think the R implementation also checks for convergence at the initial 2 steps (`y1` and `y2` in my implementation above). Is this something easily incorporated in `optimistix`'s current fixed point framework?

Let me look into the stabilisation strategy more and round back on this. I think this should be a blocker for the PR.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+import jax
+import jax.numpy as jnp
+from equinox.internal import ω
+from jaxtyping import Array, Bool, PyTree, Scalar
+
+from .._custom_types import Aux, Fn, Y
+from .._fixed_point import AbstractFixedPointSolver
+from .._misc import max_norm
+from .._solution import RESULTS
+from .._solver.fixed_point import _FixedPointState
```
johannahaffner, line 14: Nit: Let's avoid a private import here and just redeclare a `_SquarEM` state in this file.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+            alpha = self.scheme(r, v)
+            # per Varadhan and Roland (2008), set the stepsize to be at most -1
+            alpha = ω(alpha).call(lambda x: jnp.minimum(x, -1.0))
+            y_prime = (y**ω - 2.0 * alpha * r**ω + alpha**2 * v**ω).ω
+        # last stabilisation step
+        new_y, aux = fn(y_prime, args)
+        error = (y**ω - new_y**ω).ω
+        with jax.numpy_dtype_promotion("standard"):
+            scale = (self.atol + self.rtol * ω(new_y).call(jnp.abs)).ω
+            new_state = _FixedPointState(self.norm((error**ω / scale**ω).ω))
```
johannahaffner, line 144: Users might specify an absolute tolerance of zero here, in which case we could get a division by zero here, if `new_y` is zero. Even if it is not, this function won't actually diagnose convergence correctly whenever `abs(new_y) < 1`, because the relative error will be increased in such a case. 

I think we should revisit the definition of the relative error here, to make sure that it is robust.

optimistix/_solver/squarem.py
johannahaffner: Thanks, sounds good! I will also take a look at the R code.

optimistix/_solver/squarem.py
```diff
@@ -0,0 +1,182 @@
...
+    r"""Scheme 3 of SQUAREM algorithm steplength
+
+    Equation (9) of Varadhan and Roland (2008)
+    for the computation of $\alpha = \frac{\left\|r^Tr\right\|{\left\|r^Tv\right\|$.
+    We use the Frobenius norm to accommodate arrays with many axes.
+    """
+    # compute the L2 norms of r and v
+    l2_r = ω(r).call(lambda x: jnp.linalg.norm(x, ord=None))
+    l2_v = ω(v).call(lambda x: jnp.linalg.norm(x, ord=None))
+    return (-1.0 * l2_r / l2_v).ω
```
johannahaffner, line 49: I believe this would require a guard against a division by zero, which probably entails modifying the definition given in the papers.

optimistix/_solver/squarem.py
johannahaffner: Yes, that makes sense :)

optimistix/_solver/squarem.py
johannahaffner: Yes, perfect :)

johannahaffner:
This PR was closed automatically due to today's release. Feel free to reopen it whenever you want to pick this up again, I can also help rebase it.

PR #156, opened 2025-08-04
Fixed missing dtype for Newton/Chord

patrick-kidger:


benchmarks/levenberg-marquardt.py
```diff
@@ -131,6 +131,9 @@ def compute(self, state, vector, options):
             {},
         )
 
+    def conj(self, state, options):
+        assert False
+
```
patrick-kidger: (This change is just an unrelated drive-by from an update over in Lineax, needed for pre-commits to pass.)

johannahaffner:
And merged! Thanks for the fix :)

PR #157, opened 2025-08-04
Tidying up the commit history in dev

johannahaffner:
This is a tidy version of the history of `dev` that I had messed up. 

It is rebased on `main`. Before the rebase on main its diff with `upstream/dev` was nil. All tests pass, it includes all the changes from the recent PRs not yet merged into `main`, with PRs in a single commit with the exception of `DFP`, which already included four commits in the pre-mess-up version of `dev`.

Making this a PR to facilitate cross-checking :)


patrick-kidger:
Awesome. `dev` is dead, long live `dev`.

I think I'm generally in favour of keeping things around for posterity - the cost is low and it's invaluable if we mess up something here.

To that point, I recommend we create a `messy-dev` branch for the messy `dev` without rebasing on `main`. Then call this one `dev`, make sure it's rebased on `main`, and retarget the PRs you highlight on to the new `dev` (i.e. this branch) and get them in.

As for doing another release, my plan is to do this just as soon as JAX 0.7.1 comes out, what with JAX 0.7.0 causing crashes and what-not. We can add `!=0.7.0` as a version constraint to help users in this regard.

johannahaffner:
May `dev` live a happy life until its peaceful release.

Thank you for your patience with me! I pushed `messy-dev` with all the unsquashed commits from the L-BFGS PR, and this branch, renamed as `tidy-dev` so as not to confuse our future selves with the `-2`. 

Closing this PR :)

PR #169, opened 2025-09-24
Golden search minimiser

johannahaffner:
This is a simple minimisation routine that does not require gradient evaluation. Basically the minimisation cousin of `Bisection`, except that it keeps around three reference points and not two.

We use this as a subroutine in certain line searches for constrained optimisation.

### DIFF
--- a/docs/api/minimise.md
+++ b/docs/api/minimise.md
@@ -90,6 +90,11 @@ In addition to the following, note that the [Optax](https://github.com/deepmind/
 
 ---
 
+::: optimistix.GoldenSearch
+    options:
+        members:
+            - __init__
+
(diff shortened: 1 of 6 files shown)


johannahaffner:
Changes: 

- removed the function values at the initial lower and upper bound, since these are never checked. 
- removed the `_Point` module in favour of pure `Float[Array, ""]` types
- doc improvements

I did keep opt to keep tracking the middle point in the state. This is because we do need to have access to the function value at this point, and evaluating this in `init` for the first step and then carrying it forward as part of the state means we only need one evaluation of `fn` each time `step` is called. 
With `f_middle` being written into the state, I think it is more intuitive that the middle point itself is written into the state whenever `f_middle` is, too.

PR #177, opened 2025-10-24
Zoom linesearch

bagibence:
Reopening #143 as it was closed when its base branch was deleted on release.

### DIFF
--- a/optimistix/_search.py
+++ b/optimistix/_search.py
@@ -320,6 +320,8 @@ class AbstractSearch(eqx.Module, Generic[Y, _FnInfo, _FnEvalInfo, SearchState]):
     See [this documentation](./introduction.md) for more information.
     """
 
+    _needs_grad_at_y_eval: ClassVar[bool]
+
     @abc.abstractmethod
     def init(self, y: Y, f_info_struct: _FnInfo) -> SearchState:
         """Is called just once, at the very start of the entire optimisation problem.
@@ -363,7 +365,7 @@ class AbstractSearch(eqx.Module, Generic[Y, _FnInfo, _FnEvalInfo, SearchState]):
         - `f_info`: An [`optimistix.FunctionInfo`][] describing information about `f`
             evaluated at `y`, the gradient of `f` at `y`, etc.
         - `f_eval_info`: An [`optimistix.FunctionInfo`][] describing information about
-            `f` evaluated at `y`, the gradient of `f` at `y`, etc.
+            `f` evaluated at `y_eval`, the gradient of `f` at `y_eval`, etc.
         - `state`: the evolving state of the repeated searches.
 
         **Returns:**
--- a/optimistix/_solver/gradient_methods.py
+++ b/optimistix/_solver/gradient_methods.py
@@ -1,5 +1,5 @@
 from collections.abc import Callable
-from typing import Any, Generic, TypeAlias
+from typing import Any, cast, Generic, TypeAlias
 
 import equinox as eqx
 import jax
@@ -131,7 +131,9 @@ class AbstractGradientDescent(AbstractMinimiser[Y, Aux, _GradientDescentState]):
     norm: AbstractVar[Callable[[PyTree], Scalar]]
     descent: AbstractVar[AbstractDescent[Y, FunctionInfo.EvalGrad, Any]]
     search: AbstractVar[
-        AbstractSearch[Y, FunctionInfo.EvalGrad, FunctionInfo.Eval, Any]
+        AbstractSearch[
+            Y, FunctionInfo.EvalGrad, FunctionInfo.Eval | FunctionInfo.EvalGrad, Any
+        ]
     ]
 
     def init(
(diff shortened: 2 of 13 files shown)


optimistix/_solver/gradient_methods.py
```diff
@@ -170,19 +172,31 @@ def step(
         f_eval, lin_fn, aux_eval = jax.linearize(
             lambda _y: fn(_y, args), state.y_eval, has_aux=True
         )
+
+        if self.search._needs_grad_at_y_eval:
```
patrick-kidger, line 176: FYI this is verboten, accessing a private field.

optimistix/_solver/gradient_methods.py
```diff
@@ -170,19 +172,31 @@ def step(
...
 
         def accepted(descent_state):
-            grad = lin_to_grad(lin_fn, state.y_eval, autodiff_mode=autodiff_mode)
+            nonlocal f_eval_info
+
+            if not self.search._needs_grad_at_y_eval:
+                grad = lin_to_grad(lin_fn, state.y_eval, autodiff_mode=autodiff_mode)
+                f_eval_info = FunctionInfo.EvalGrad(f_eval, grad)
+
+            f_eval_info = cast(FunctionInfo.EvalGrad, f_eval_info)
```
patrick-kidger: Better to `assert`, which checks this at runtime as well as statically.

optimistix/_search.py
```diff
@@ -320,6 +320,8 @@ class AbstractSearch(eqx.Module, Generic[Y, _FnInfo, _FnEvalInfo, SearchState]):
     See [this documentation](./introduction.md) for more information.
     """
 
+    _needs_grad_at_y_eval: ClassVar[bool]
```
patrick-kidger, line 323: Should this be an `AbstractClassVar`?

If so then FYI this is a breaking change, since any other `AbstractSearch` subclasses out there will need to provide it.

optimistix/_solver/gradient_methods.py
```diff
@@ -170,19 +172,31 @@ def step(
         f_eval, lin_fn, aux_eval = jax.linearize(
             lambda _y: fn(_y, args), state.y_eval, has_aux=True
         )
+
+        if self.search._needs_grad_at_y_eval:
+            grad = lin_to_grad(lin_fn, state.y_eval, autodiff_mode)
+            f_eval_info = FunctionInfo.EvalGrad(f_eval, grad)
+        else:
+            f_eval_info = FunctionInfo.Eval(f_eval)
```
patrick-kidger, line 180: What if we need some other flavor of `FunctionInfo` later? I think picking a boolean flag here may paint us into a corner in the future. Perhaps a more general approach is required.

optimistix/_solver/quasi_newton.py
```diff
@@ -210,29 +207,40 @@ def step(
         f_eval, lin_fn, aux_eval = jax.linearize(
             lambda _y: fn(_y, args), state.y_eval, has_aux=True
         )
+
+        if self.search._needs_grad_at_y_eval:
```
patrick-kidger, line 211: I am really coming to dislike the code duplication we have between these routines >.<

(Totally my fault a few years ago, no idea what the the solution is.)

optimistix/_solver/gradient_methods.py
bagibence: How about just making it public? It's only accessed from minimisers.

optimistix/_search.py
bagibence: Not sure. Zoom is the only one that uses it now, but I also added it to all other searches.

How about just defaulting to `False` in `AbstractSearch` and only overwriting it in `Zoom`?

optimistix/_solver/gradient_methods.py
bagibence: Do you mean that it might not just be `grad` that's required by a future search?

Maybe the search could have a method to set up `f_eval_info`? Or give a list of minimum required info fields that need to be filled in here?
Then in `accepted`, instead of the flag, branch based on what fields `f_eval_info` already has? So here it could be `if not hasattr(f_eval_info, "grad")`.

optimistix/_solver/gradient_methods.py
bagibence: Changed to `assert isinstance(f_eval_info, FunctionInfo.EvalGrad)`

optimistix/_solver/gradient_methods.py
patrick-kidger: Definitely 'no' to using `hasattr`, which I'm not a fan of. I generally argue that it should be readable what types that a piece of code is expecting, which means that one should use `isinstance` instead of `hasattr` in order to determine what is available :)

But yes, I'm thinking that a future search may need more than grad – in general any kind of `FunctionInfo`. Perhaps something like `info_needed_at_y_eval: AbstractClassVar[FunctionInfo]`? @johannahaffner may have stronger opinions than I do here though.

bagibence:
Please excuse me for the delayed reply.

> I'm afraid this might be something we're not able to work on for a while.

I'm sad to hear that, but of course completely understand that you don't have the capacity for this now.

> In the interests of unblocking you, I think the best we can aim for is to land just those pieces – something as small and standalone as possible (i.e. as easy to review as possible) – so as to enable that third-party use-case.

Absolutely, let's aim for that for now. It would be nice to have the option to use this without having to depend on a fork.

> Adding a short thing: I have experimented with passing grad_at_y_eval unconditionally to all searches and ran benchmarks for this and it resulted in zero performance pessimisation.

That sounds great! How about I just make this adjustment, run the benchmarks suggested by @johannahaffner and send it as a small PR?
Or do you guys maybe have another design in mind?

johannahaffner:
> > Adding a short thing: I have experimented with passing grad_at_y_eval unconditionally to all searches and ran benchmarks for this and it resulted in zero performance pessimisation.

> That sounds great! How about I just make this adjustment, run the benchmarks suggested by @johannahaffner and send it as a small PR? Or do you guys maybe have another design in mind?

I think that would be good! Having access to the gradients in the searches is quite common and something we'll also need to land support for constrained optimisation.

BalzaniEdoardo:
#233 @bagibence there is still interest in the zoom search

PR #183, opened 2025-10-31
add strong wolfe condition

jaeminoh:
Hi, I've implemented the strong Wolfe conditions for the backtracking line search. 

However, one potential issue related to the class of QuasiNewton's methods is that the sufficient decrease condition

$$|\nabla f(y_{k+1}) \cdot (y_{k+1} - y_k)| \le c | \nabla f(y_k) \cdot (y_{k+1} - y_k) |$$

requires $\nabla f(y_{k+1})$. Currently, the `step` method of the `AbstractQuasiNewton` class does not pass such information ([line 213--220](https://github.com/patrick-kidger/optimistix/blob/5b1fa3005993f4f098e9532de9d21e5513a6d550/optimistix/_solver/quasi_newton.py#L213C1-L220C10)): 

```python
step_size, accept, search_result, search_state = self.search.step(
            state.first_step,
            y,
            state.y_eval,
            state.f_info,
            FunctionInfo.Eval(f_eval),
            state.search_state,
        )
```

### DIFF
--- a/optimistix/_solver/backtracking.py
+++ b/optimistix/_solver/backtracking.py
@@ -110,3 +110,112 @@ BacktrackingArmijo.__init__.__doc__ = """**Arguments:**
 - `step_init`: The first `step_size` the backtracking algorithm will
     try. Must be greater than 0.
 """
+
+
+class BacktrackingStrongWolfe(
+    AbstractSearch[Y, _FnInfo, _FnEvalInfo, _BacktrackingState]
+):
+    """Perform a backtracking line search satisfying the strong Wolfe conditions."""
+
+    decrease_factor: ScalarLike = 0.5
+    slope: ScalarLike = 1e-4
+    step_init: ScalarLike = 1.0
+    c: ScalarLike = 0.9
+
+    def __post_init__(self):
+        self.decrease_factor = eqx.error_if(
+            self.decrease_factor,
+            (self.decrease_factor <= 0)  # pyright: ignore
+            | (self.decrease_factor >= 1),  # pyright: ignore
+            "`BacktrackingStrongWolfe(decrease_factor=...)` must be between 0 and 1.",
+        )
+        self.slope = eqx.error_if(
+            self.slope,
+            (self.slope <= 0) | (self.slope >= 1),  # pyright: ignore
+            "`BacktrackingStrongWolfe(slope=...)` must be between 0 and 1.",
+        )
+        self.step_init = eqx.error_if(
+            self.step_init,
+            self.step_init <= 0,  # pyright: ignore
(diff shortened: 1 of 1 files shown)


optimistix/_solver/backtracking.py
```diff
@@ -110,3 +110,112 @@ def step(
...
+        y: Y,
+        y_eval: Y,
+        f_info: _FnInfo,
+        f_eval_info: _FnEvalInfo,
+        state: _BacktrackingState,
+    ) -> tuple[Scalar, Bool[Array, ""], RESULTS, _BacktrackingState]:
+        if not isinstance(
+            f_info,
+            (
+                FunctionInfo.EvalGrad,
```
johannahaffner, line 164: You can define a union type outside the search, e.g. with `_FnInfo = FunctionInfo.EvalGrad | FunctionInfo.EvalGradHessian | ... `

optimistix/_solver/backtracking.py
```diff
@@ -110,3 +110,112 @@ def step(
...
+BacktrackingStrongWolfe.__init__.__doc__ = """**Arguments:**
+
+- `decrease_factor`: The rate at which to backtrack, i.e.
+    `next_stepsize = decrease_factor * current_stepsize`. Must be between 0 and 1.
+- `slope`: The slope of of the linear approximation to
+    `f` that the backtracking algorithm must exceed to terminate. Larger
+    means stricter termination criteria. Must be between 0 and 1.
+- `step_init`: The first `step_size` the backtracking algorithm will
+    try. Must be greater than 0.
+- `c`: The parameter for the strong Wolfe condition. Must be between 0 and 1.
```
johannahaffner, line 220: Can we find a more descriptive name for this parameter?

johannahaffner:
> I have implemented this because this was not in `optimstix` but was in `jaxopt`. I’m not sure how to proceed, but perhaps a small refactoring to reduce code duplication would be better?

This one? https://jaxopt.github.io/stable/_autosummary/jaxopt.HagerZhangLineSearch.html? 

> I’m not sure how to proceed, but perhaps a small refactoring to reduce code duplication would be better?

Do you mean code duplication between Optimistix and JAXopt? If you're referring to this PR - without a means to locate a promising interval, checking the conditions isn't enough. If you'd like to keep experimenting, then I suggest adding your search to custom solvers defined here

https://github.com/patrick-kidger/optimistix/blob/23b99d29d3274a65d58e9b9a3d2fca79fc65666f/tests/helpers.py#L84

and running the tests for these - currently your search is completely untested. If you'd like to keep working on making a `HagerZhang` line search instead, then this is not a small refactoring, but rather a quite substantial technical task. If that is not for you, that is fine too - you can just close this PR as well.

PR #184, opened 2025-11-01
Routine to identify the generalised Cauchy point

johannahaffner:
This adds a small module with a routine that identifies the generalised Cauchy point, the first local minimiser along the piecewise linear path defined by the gradient, projected on the box defined by bound constraints. 

This is an essential ingredient in `BFGS-B` and `L-BFGS-B`, and implemented as described in the original publication: https://epubs.siam.org/doi/10.1137/0916069
As with many constrained methods, there are lots of edge cases to think of + get right here, so I added a bunch of tests to cycle through the ones I could think of. 

I think that the `cauchy_point` module can be merged into `main`: I would not add it to the public API, and the change is purely additive.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1294,6 +1294,10 @@ minimise_bounded_with_local_minima = (
 )
 
 
+# Cauchy point tests: these all use a simple paraboloid centered around the origin as
+# the objective function. This means that the gradient evaluated at any of the
+# following test points is equal to the reflection of this point about the origin, and
+# the bounds determine where the Cauchy point then lies.
 def _wrapped_paraboloid(y):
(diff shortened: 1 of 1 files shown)


johannahaffner:
Tagging @cjchristopher - this is the routine to identify the generalised Cauchy point, required for `BFGS-B` and `L-BFGS-B`. I'm pretty confident that it does the right thing, but another set of eyes would be great!

You're right about the strong Wolfe condition in the line search proposed in the Byrd et al. paper. I haven't checked how they modify the line search to ensure that it stays in the feasible region, though. Right now just trying to divide and conquer the pile of code I already wrote to make constrained optimisation happen in Optimistix :)

Issue #185, opened 2025-11-05
optimistix 0.0.11 causes jax to initialize on import


Findus23:
Hi,

I updated to optimistix 0.0.11 and realized that some of my code doesn't work the same anymore:

Normally jax initializes the backend and xla when the first array is created or something like `jax.devices()`  is called. So many global settings and environment variables have to be set before. 
Importing most jax libraries doesn't do any of these things, so they are generally safe (but I agree that in theory one should delay everything after all settings are set).
But importing optimistix 0.0.11 does initialize jax:
```python
import os
import jax
from jax._src import xla_bridge as xb
print("jax backend:", xb._default_backend)
import optimistix
print("jax backend:", xb._default_backend)
jax.config.update('jax_platform_name', 'cpu')
os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=4"
print(jax.devices())
```
returns
```text
jax backend: None
jax backend: <jaxlib._jax.Client object at 0x721da5f2f2a0>
[CudaDevice(id=0)]
```

So importing optimistix (or importing a module that somewhere imports diffrax which imports optimistix) creates one jax object here:
https://github.com/patrick-kidger/optimistix/blob/5b1fa3005993f4f098e9532de9d21e5513a6d550/optimistix/_misc.py#L138

This isn't a big issue (as one in theory should never import jax-related code before having all options set), but it is easy to stumble into this unless one always check if `xla_bridge._default_backend` already exists when setting XLA options).

johannahaffner:
Thanks for the report! I can look at this next week - off the top of my head I'm not sure where this might come from.

Findus23:
Thanks, #186 looks like a nice workaround that should fix this issue.

PR #186, opened 2025-11-05
Avoid triggering JAX initialization on import

johannahaffner:
Fixes https://github.com/patrick-kidger/optimistix/issues/185#issuecomment-3492999481. 

The presence of a keyword-argument with an array-valued default caused JAX initialisation on import. Setting the value of the default argument to None and then setting it to `jnp.array(0.0)` at compile time avoids this.

### DIFF
--- a/optimistix/_misc.py
+++ b/optimistix/_misc.py
@@ -135,7 +135,7 @@ def feasible_step_length(
     lower_bound: PyTree[ArrayLike, " T"],  # pyright: ignore[reportUndefinedVariable]
     upper_bound: PyTree[ArrayLike, " T"],  # pyright: ignore[reportUndefinedVariable]
     *,
-    offset: ScalarLike = jnp.array(0.0),
+    offset: ScalarLike | None = None,
 ) -> PyTree[ArrayLike, " T"]:  # pyright: ignore[reportUndefinedVariable]
     """Returns the maximum feasible step length for any current value, its bounds, and a
     proposed step, as a value for each element of the PyTree.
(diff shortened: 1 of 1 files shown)


optimistix/_misc.py
```diff
@@ -135,7 +135,7 @@ def feasible_step_length(
     lower_bound: PyTree[ArrayLike, " T"],  # pyright: ignore[reportUndefinedVariable]
     upper_bound: PyTree[ArrayLike, " T"],  # pyright: ignore[reportUndefinedVariable]
     *,
-    offset: ScalarLike = jnp.array(0.0),
+    offset: ScalarLike | None = None,
```
jpbrodrick89, line 138: Just working on something similar and thought I'd throw some other ideas around in case you're looking for better solutions.

Converter:
```suggestion
    offset: ScalarLike = eqx.field(converter=jnp.asarray, default=0.0)
```


or default factory (not really necessary here but could be useful in other contexts)
```suggestion
    offset: ScalarLike = eqx.field(default_factory = lambda: jnp.array(0.0)) 
```

johannahaffner: Interesting! A field that is not a dataclass field, this hadn't occurred to me. 

What do you prefer about this solution, other than not requiring the `if offset is None` condition?

jpbrodrick89: Sorry, I was having this problem (or at least protecting against it) in an equinox Module late at night and I just thought that you might have been having this problem with an equinox Module, I didn't double-check. I don't think this will work for a simple function, sorry! 😅

jpbrodrick89: I would resolve this, but it doesn't seem possible for pull requests that are already merged!

johannahaffner: No worries at all! Just leave it up and have a nice day :)

Issue #187, opened 2025-11-07
Scale-invariant Levenberg-Marquardt


michael-0brien:
Hi `optimistix` developers! I was hoping to start a discussion about improving `optimistix` support for working in "unnormalized" parameter spaces. Let me explain what I mean by this.

I come from a scientific domain where the task is to infer the parameters of a physical model from messy experimental data (think gradient-based optimization of a loss function with respect to the parameters of a forward model for the data). In these tasks, it is very important to define your optimizer starting with some heuristics about your parameter space.

Arguably the most important / most common thing needed is to define what a "unit" step size is. Physical parameters vary on vastly different scales; one reason is due to the fact that parameters have units, another is simply due to the fact that loss function geometry is non-trivial. This is related to the success of the MCMC sampler implemented in [`emcee`](https://github.com/dfm/emcee/tree/main) in the physical sciences, which does not care about these factors due to its affine invariance property. 

I was generally wondering your thoughts on this, and also hoping you may accept a PR adding support for defining "unit" step size in the appropriate cases. In my [fork](https://github.com/michael-0brien/optimistix) of `optimistix`, I've implemented three things that are useful for me in this regard:

- `ClassicalTrustRegion`s should allow users to specify an `initial_step_size`, rather than always defaulting to `1.0`. In the coming weeks I can try to put together a MRE why this is important, but I believe in unnormalized parameter spaces convergence should be much faster if this can be tuned.
- `DampedNewtonDescent` should define the Levenberg-Marquardt parameter as `unit_step_size / step_size`, rather than `1 / step_size`. `1` may not be meaningful in the case I describe.
- Solvers should allow users to optionally pass their respective `AbstractSearch` instances. I want to have the option to change `ClassicalTrustRegion` parameters. This also seems appropriate for `BacktrackingArmijo` in the BFGS solvers.

These are the changes that would be helpful for my work, but there may other changes that would add support for the general theme I am describing.

Note: I understand that there are other ways of handling the cases I describe, such as applying parameter rescalings directly to my model. I believe what I'm proposing is closely related to the ability to tune "learning_rate" in `optax`-like optimizers, which is API that many people are accustomed to (and for me is much easier syntaxwise than applying rescalings). It would be good to give users some flexibility in this regard.

michael-0brien:
I was reading up on some literature on this problem and have been looking through `scipy`’s implementation of LM. It is based on the work of 
[...]

PR #188, opened 2025-11-09
Specify initial step size in `ClassicalTrustRegion`

michael-0brien:
See discussion in #187 about specifying an initial step size in the `ClassicalTrustRegion`. I've added the corresponding parameters to `Dogleg`, `LevenbergMarquardt`, and `IndirectLevenbergMarquardt`, and I've named things according to the language used in the docstrings.


michael-0brien:
Gotcha, based on our conversation this makes sense. I'll close this PR in favor of another one at some point. If you happen to get time before I do, then feel free to go ahead and make the change.

PR #189, opened 2025-11-11
Scale-invariant levenberg marquardt

michael-0brien:
Here's my draft of the `AbstractDescent` as described in #187. I also have the inclusion of an initial step size to the `ClassicalTrustRegion` here, but we can move this to another PR if desired.

For making scale-invariance the default implementation, we will have to discuss how to strike a balance between scale-invariance and robustness to bad initialization. The jist is that if the initial hessian estimate is bad, the algorithm loses parameter sensitivity.

The scipy version involves taking the scaling matrix `D` to be the max diagonal of the hessian so far encountered. They probably have other tricks, but this alone is not robust to bad initialization. I could imagine other cases where one would not want the max, perhaps one would be interested in the average instead. Regardless, we will need some modularity for implementing the scaling operator.

The paper I shared in #187 has some advice for choices of `D`, such as adding small const diagonal offset to prevent becoming too insensitive to parameters. In my work, I'm having success normalizing the trace of `D` to be equal to the identity. That way the LM param too carries a sense of scale and I can use it to bias the algorithm to gradient-like or newton-like steps in early iterations. `D` then maintains relative scale between parameters. I'm not suggesting this should be the default, but some food for thought.

### DIFF
--- a/optimistix/_solver/levenberg_marquardt.py
+++ b/optimistix/_solver/levenberg_marquardt.py
@@ -48,7 +48,7 @@ def damped_newton_step(
     f_info: FunctionInfo.EvalGradHessian | FunctionInfo.ResidualJac,
     linear_solver: lx.AbstractLinearSolver,
     *,
-    scaling_update: _ScalingUpdate | None,
+    scaling_operator: lx.AbstractLinearOperator | None,
 ) -> tuple[PyTree[Array], RESULTS]:
     """Compute a damped Newton step.
 
@@ -73,16 +73,14 @@ def damped_newton_step(
     lm_param = jnp.where(pred, 1 / safe_step_size, jnp.finfo(step_size.dtype).max)
     lm_param = cast(Array, lm_param)
     if isinstance(f_info, FunctionInfo.EvalGradHessian):
-        if scaling_update is None:
+        if scaling_operator is None:
             scaling_operator = lx.IdentityLinearOperator(f_info.hessian.in_structure())
-        else:
-            scaling_operator = scaling_update(f_info.hessian)
         operator = f_info.hessian + lm_param * scaling_operator
         vector = f_info.grad
         if lx.is_positive_semidefinite(f_info.hessian):
             operator = lx.TaggedLinearOperator(operator, lx.positive_semidefinite_tag)
     elif isinstance(f_info, FunctionInfo.ResidualJac):
-        if scaling_update is None:
+        if scaling_operator is None:
             y_structure = f_info.jac.in_structure()
             operator = lx.FunctionLinearOperator(
                 _Damped(f_info.jac, lm_param), y_structure
@@ -90,7 +88,6 @@ def damped_newton_step(
             vector = (f_info.residual, tree_full_like(y_structure, 0))
         else:
             hessian = f_info.jac.T @ f_info.jac
-            scaling_operator = scaling_update(hessian)
             operator = lx.TaggedLinearOperator(
                 hessian + lm_param * scaling_operator, lx.positive_semidefinite_tag
             )
@@ -155,7 +152,7 @@ class DampedNewtonDescent(
         self, step_size: Scalar, state: _DampedNewtonDescentState
     ) -> tuple[Y, RESULTS]:
         sol_value, result = damped_newton_step(
-            step_size, state.f_info, self.linear_solver, scaling_update=None
+            step_size, state.f_info, self.linear_solver, scaling_operator=None
         )
         y_diff = (-(sol_value**ω)).ω
         return y_diff, result
@@ -242,7 +239,7 @@ class IndirectDampedNewtonDescent(
 
         def comparison_fn(lambda_i: Scalar, _):
             step, _ = damped_newton_step(
-                1 / lambda_i, state.f_info, self.linear_solver, scaling_update=None
+                1 / lambda_i, state.f_info, self.linear_solver, scaling_operator=None
             )
             return self.trust_region_norm(step) - scaled_step_size
 
@@ -257,7 +254,7 @@ class IndirectDampedNewtonDescent(
                 throw=False,
             ).value
             y_diff, result = damped_newton_step(
-                1 / lambda_out, state.f_info, self.linear_solver, scaling_update=None
+                1 / lambda_out, state.f_info, self.linear_solver, scaling_operator=None
             )
             return y_diff, result
 
@@ -361,14 +358,19 @@ class ScaledDampedNewtonDescent(
         state: _ScaledDampedNewtonDescentState,
     ):
         del y
-        return _ScaledDampedNewtonDescentState(f_info, state.scaling_operator)
+        scaling_operator = self.update_scaling_fn(
+            f_info.hessian, state.scaling_operator
+        )
+        return _ScaledDampedNewtonDescentState(f_info, scaling_operator)
 
     def step(
         self, step_size: Scalar, state: _ScaledDampedNewtonDescentState
     ) -> tuple[Y, RESULTS]:
-        scaling_update = _ScalingUpdate(self.update_scaling_fn, state.scaling_operator)
         sol_value, result = damped_newton_step(
-            step_size, state.f_info, self.linear_solver, scaling_update=scaling_update
+            step_size,
+            state.f_info,
+            self.linear_solver,
+            scaling_operator=state.scaling_operator,
         )
         y_diff = (-(sol_value**ω)).ω
         return y_diff, result


michael-0brien:
Two things added to public API so far:

- `ScaledDampedNewtonDescent`: The `AbstractDescent` for direct scale-invariant LM. This keeps track of the scaling operator and allows a user to pass a callable `update_scaling_fn(hessian: lx.AbstractLinearOperator, scaling_operator: lx.DiagonalLinearOperator) -> lx.DiagonalLinearOperator`.
- `max_diagonal_scaling_update`: The "scipy" scaling update of taking the maximum diagonal of the hessian so far encountered. I thought that exposing scaling update functions to public API could follow functions for computing beta in non-linear CG.

Not sure if this is fully the way to go but should get the ball rolling! I've also written up some documentation and added to tests.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -31,42 +37,79 @@ def __call__(self, y: PyTree[Array]):
...
+
+    def __call__(self, hessian: lx.AbstractLinearOperator) -> lx.DiagonalLinearOperator:
+        return self.update_scaling_fn(hessian, self.scaling_operator)
+
+
+def max_diagonal_scaling_update(
+    hessian: lx.AbstractLinearOperator, scaling_operator: lx.DiagonalLinearOperator
+) -> lx.DiagonalLinearOperator:
+    """Update the matrix `D` that controls the relative scaling of each
+    parameter based on the procedure described by More (1977).
```
johannahaffner: Let's avoid all confusion and just point out that the full reference is given below :)

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -196,7 +239,10 @@ def query(
         newton, result = newton_step(f_info, self.linear_solver)
         newton_norm = self.trust_region_norm(newton)
         return _IndirectDampedNewtonDescentState(
-            f_info=f_info, newton=newton, newton_norm=newton_norm, result=result
+            f_info=f_info,
```
johannahaffner: This looks like a spurious formatting change - best to remove it, I think.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -31,42 +37,79 @@ def __call__(self, y: PyTree[Array]):
         return residual, damped
 
 
+class _ScalingUpdate(eqx.Module):
+    update_scaling_fn: UpdateScalingFn
+    scaling_operator: lx.DiagonalLinearOperator
+
+    def __call__(self, hessian: lx.AbstractLinearOperator) -> lx.DiagonalLinearOperator:
+        return self.update_scaling_fn(hessian, self.scaling_operator)
```
johannahaffner: I don't think this will work if `scaling_update` is `None`, which we do implicitly allow since this is one of the things we check in `damped_newton_step` below. 

I'll leave another comment at the call site.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -253,6 +299,83 @@ def accept_newton():
...
+
+    - J. J. More, “The Levenberg-Marquardt Algorithm: Implementation and Theory,”
+    Numerical Analysis, ed. G. A. Watson, Lecture Notes in Mathematics 630,
+    Springer Verlag, pp. 105-116, 1977.
+    """  # noqa: E501
+
+    # Will probably resolve to either Cholesky (for minimisation problems) or
+    # QR (for least-squares problems).
+    linear_solver: lx.AbstractLinearSolver = lx.AutoLinearSolver(well_posed=None)
+    update_scaling_fn: UpdateScalingFn = max_diagonal_scaling_update
```
johannahaffner: I think we should consider the type of the scaling function we want to support here. How about we make it a `Callable` that can do anything as long as it takes two Lineax operators and returns another Lineax operator? This will make it easy for anyone wanting to implement a custom thing.

johannahaffner (commented):
This looks very good to me! (I thought I had more comments but upon going over this again they resolved themselves.) 

The main thing is that I think the function updating the scaling factor should become more flexible, and just be a `Callable` rather than a custom type. Perhaps `Callable[[lx.AbstractLinearOperator, lx.AbstractLinearOperator], lx.AbstractLinearOperator]` would work?

As you've suggested above, let's put the initial step size attribute for the searches in a different PR.

johannahaffner:
> The jist is that if the initial hessian estimate is bad, the algorithm loses parameter sensitivity.

How does this play out with different flavours of Hessian approximations? A `BFGS` type Hessian approximation that starts out with an identity should basically implement gradient descent for the first few steps. What happens with a residual Jacobian far from the optimum?

> I could imagine other cases where one would not want the max, perhaps one would be interested in the average instead. 

>  In my work, I'm having success normalizing the trace of `D` to be equal to the identity. That way the LM param too carries a sense of scale and I can use it to bias the algorithm to gradient-like or newton-like steps in early iterations. 

> Regardless, we will need some modularity for implementing the scaling operator.

I think making the type of the update function more flexible should address all of the above, right? This does not answer the question of what the best default is, but we could start out with the maximum value, matching `scipy`, and make a change if we see something else that performs much better. In any case, we should write in the documentation that this can be highly problem-specific. 

We could also provide several scaling functions, similar to how we have implemented this for the `NonlinearCG` family, which may be used with different methods: https://docs.kidger.site/optimistix/api/minimise/#optimistix.NonlinearCG (that are also just restricted to implement a specific signature).

michael-0brien:
Thanks for the comments! I’ll work on finalizing this PR. Two questions in finalizing: 

1. What API would we like for the `LevenbergMarquardt` class? Should the scaled update become the default and if so, how? Didn’t want to touch this part since I’m not sure what is desired for backwards compatibility. Somehow it should be clear that the new approach may break for large systems since it requires materializing the hessian. 
2. Would you prefer I revert my commit adding the `initial_step_size` to the search method, or would you want to cherry pick my commit for the scaled descent? If the latter, I’ll add another commit addressing your comments.

> > The jist is that if the initial hessian estimate is bad, the algorithm loses parameter sensitivity.
> 
> How does this play out with different flavours of Hessian approximations? A `BFGS` type Hessian approximation that starts out with an identity should basically implement gradient descent for the first few steps. What happens with a residual Jacobian far from the optimum?

Good question, I need to read up on this. The work I’ve seen that studies this is always in context of the usual residual jacobian LM. I think this is still true in this case, which is why I’m curious if scipy does anything else under the hood besides what’s implemented in the 1977 reference they cite. 

I am using a BFGS hessian of course, so I just bias the algorithm in early iterations to be gradient-like. I do this by choosing the initial LM param to be large, and also by using a custom search that has the behavior of `LinearTrustRegion` before some `hessian_burn_in` number of steps, and `ClassicalTrustRegion` after that.

> > I could imagine other cases where one would not want the max, perhaps one would be interested in the average instead.
> 
> > In my work, I'm having success normalizing the trace of `D` to be equal to the identity. That way the LM param too carries a sense of scale and I can use it to bias the algorithm to gradient-like or newton-like steps in early iterations.
> 
> > Regardless, we will need some modularity for implementing the scaling operator.
> 
> I think making the type of the update function more flexible should address all of the above, right? This does not answer the question of what the best default is, but we could start out with the maximum value, matching `scipy`, and make a change if we see something else that performs much better. In any case, we should write in the documentation that this can be highly problem-specific.
> 
> We could also provide several scaling functions, similar to how we have implemented this for the `NonlinearCG` family, which may be used with different methods: https://docs.kidger.site/optimistix/api/minimise/#optimistix.NonlinearCG (that are also just restricted to implement a specific signature).

Yes, I think so. Going with the scipy version and providing flexibility is a good plan!

optimistix/_solver/levenberg_marquardt.py
michael-0brien: So are you suggesting this should be changed to `Callable[[lx.AbstractLinearOperator, lx.AbstractLinearOperator], lx.AbstractLinearOperator]`?

This sounds good to me but will just need to add a `scaling_operator = cast(lx.DiagonalLinearOperator, scaling_operator)` to make pyright happy in the `update_scaling_fn`. Note that `descent.init` initializes as:

`scaling_operator = lx.DiagonalLinearOperator(tree_full_like(y, -jnp.inf, allow_static=True))`

johannahaffner:
> Comments should be addressed. Added the new `ScaledLevenbergMarquardt`, `max_diagonal_scaling_update`, and `ScaledDampedNewtonDescent` to the docs.

Great, will take a look tomorrow!

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -253,6 +280,102 @@ def accept_newton():
 """
 
 
+def max_diagonal_scaling_update(
+    hessian: lx.AbstractLinearOperator, scaling_operator: lx.AbstractLinearOperator
+) -> lx.AbstractLinearOperator:
+    """Update the scaling matrix `D` as the maximum diagonal of the
```
johannahaffner: Wouldn't it be desirable to regularise the Hessian less stringently as the optimisation progresses? If we keep the maximum value around forever I would expect convergence to slow down - the regulariser will be too dominant, and the system will become too gradient-like for fast convergence.

docs/api/least_squares.md
```diff
@@ -53,7 +53,7 @@
         members:
             - __init__
 
-[`optimistix.ScaledLevenbergMarquardt`][] supports modularity in updating the scaling matrix `D` via the `update_scaling_fn` argument. The default method is exposed to public API and is documented below.
+[`optimistix.ScaledLevenbergMarquardt`][] supports modularity in updating the scaling operator via the `update_scaling_fn` argument. The default method is exposed to public API and is documented below.
```
johannahaffner: "The default scaling method uses the maximum value along the diagonal of the Hessian approximation." Rest is self-explanatory.

optimistix/_solver/levenberg_marquardt.py
johannahaffner: Ok I have more questions about this one. In the [More 1977](https://scispace.com/pdf/levenberg-marquardt-algorithm-implementation-and-theory-9p7qhokmlq.pdf) paper, equation (6.3) does seem to point to a related concept. Here we take whatever is larger between 

- the value on the diagonal of the Hessian approximation
- the norm of the respective column in the Jacobian

in an element-wise fashion, but going back just one iteration. In other words, we would not accumulate large values throughout the optimisation. What is your take on this section of the paper?

`max_diagonal_scaling_update` should also have a reference, I think, even if this ends up being a bit repetitive given that we have references elsewhere. As soon as we'll have more than one of these scaling functions in the public API the correct reference won't be immediately obvious anymore.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -31,42 +32,68 @@ def __call__(self, y: PyTree[Array]):
...
-    For a minimisation problem, this means solving `(Hess + λI)^{-1} grad`.
+    For a minimisation problem, this means solving `(Hess + λD)^{-1} grad`.
+    `D` is typically chosen as diagonal matrix whose elements preserve
+    the scale-invariance of the usual Newton step `Hess^{-1} grad`.
 
     In the (nonlinear) least-squares case, for which the minimisation objective
     is given by `0.5*residual^2`, then we know that `grad=J^T residual`, and we make
-    the Gauss--Newton approximation `Hess ~ J^T J`. This reduces the above to
-    solving the (linear) least-squares problem
+    the Gauss--Newton approximation `Hess ~ J^T J`. When `D = I` this reduces the
```
johannahaffner: "When `D = I`, then we do not need to access the diagonal of `J^T J` and can reduce the above to...".

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -253,6 +280,102 @@ def accept_newton():
...
+):
+    """The scale-invariant damped Newton (Levenberg--Marquardt) descent.
+
+    Rather than compute updates by solving `(Hess + λI)^{-1} grad` as in
+    [`optimistix.DampedNewtonDescent`][], update based on `(Hess + λ D)^{-1} grad`,
+    where `D` controls the relative scaling of each parameter.
+
+    Taking `D = I` is the original procedure proposed by Levenberg, whereas
+    More (1977) describes taking `D` as the maximum diagonal of the hessian
+    so far encountered. This procedure is used in [`scipy.least_squares`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html)
```
johannahaffner: See comments above - I think the values are not actually accumulated throughout the whole optimisation procedure. Just leaving this note here so we don't forget to update the docstring :) 

Given that we don't actually know what MINPACK does (I'm confused by the QR factorisation taking place before any scaling does, for instance), let's drop the reference to Scipy here, move the More paper reference to the implementation of the max diagonal scaling method only.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -253,6 +280,102 @@ def accept_newton():
...
+            step_size, state.f_info, self.linear_solver, scaling_update=scaling_update
+        )
+        y_diff = (-(sol_value**ω)).ω
+        return y_diff, result
+
+
+ScaledDampedNewtonDescent.__init__.__doc__ = """**Arguments:**
+
+- `linear_solver`: The linear solver used to compute the Newton step.
+- `update_scaling_fn`: A function with signature `fn(hessian, scaling_operator)`,
```
johannahaffner: "`update_scaling_fn`: A function computing the scaling matrix `D`. Defaults to [`optimistix.max_diagonal_scaling_update`][]. `ScaledLevenbergMarquardt` supports custom scaling updates - any function implementing such an update should... " and then explain the signature.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -378,3 +501,74 @@ def __init__(
     entries are `step`, `loss`, `accepted`, `step_size`, `y`. For example
     `verbose=frozenset({"loss", "step_size"})`.
 """
+
+
+class ScaledLevenbergMarquardt(AbstractGaussNewton[Y, Out, Aux]):
+    """A Levenberg--Marquardt method invariant to parameter rescaling.
+
+    This is a variant of [`optimistix.LevenbergMarquardt`][], which uses
+    the hessian to estimate how to preserve scale-invariance when applying
```
johannahaffner: hessian -> Hessian approximation

johannahaffner (commented):
Looks good! I have some nits, some documentation requests and a question about the implementation of the diagonal scaling update - if I read the More paper correctly then they propose something different there which does not accumulate values throughout the optimisation, but rather only between successive iterations.

optimistix/_solver/levenberg_marquardt.py
michael-0brien: My interpretation of what scipy does, and the interpretation of the More paper come from this reference: https://arxiv.org/abs/1201.5885

Perhaps should not have taken this too much at face value. I’m not quite sure what the right way to go is. But if it’s true that scipy does indeed use the max diagonal so far encountered, even if this is not what is described in More, it might be the right default to go with? Points taken though.

Based on this reference do you think we should investigate scipy further or implement what is in More? If the latter, let me know what you think the best approach would be.

michael-0brien: Reading through the More paper again, I think what is described (Eq 6.3) is indeed what we’ve implemented. What makes you think the maximum is with respect to current and previous iterate?

johannahaffner: In the More paper - I interpreted the index $k-1$ as the value of the diagonal element in `Jac^T Jac` from the previous iteration, but looking at it again I can also see your interpretation. Since we're defining a recursion it actually makes a lot of sense to define it the way you did. 

What I find very surprising is that it is sensible to keep the largest value _thus far encountered_ for the entirety of the optimisation procedure. That will accumulate information from the start of the optimisation to the end, and I don't find that intuitive.

I'll take another look at the Scipy code, I'm still a bit hesitant to claim equivalence.

optimistix/_solver/levenberg_marquardt.py
```diff
@@ -31,42 +32,68 @@ def __call__(self, y: PyTree[Array]):
...
+    def __call__(self, hessian: lx.AbstractLinearOperator) -> lx.AbstractLinearOperator:
+        return self.update_scaling_fn(hessian, self.scaling_operator)
+
+
 def damped_newton_step(
     step_size: Scalar,
     f_info: FunctionInfo.EvalGradHessian | FunctionInfo.ResidualJac,
     linear_solver: lx.AbstractLinearSolver,
+    *,
+    scaling_update: _ScalingUpdate | None = None,
```
johannahaffner: Avoid the default argument here, and just keep it a mandatory keyword argument. This forces `DampedNewtonDescent` to specify that it does not use a `scaling_update`, which is cleaner and makes our internal code more self-documented.

optimistix/_solver/levenberg_marquardt.py
michael-0brien: Agreed, I see your point. This is worth looking into further, I’m also hesitant to claim equivalence. The [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html) state that the LM implementation is “based on the paper [JJMore]” and has “a lot of smart tricks.” I am not sure if this implies there are tricks beyond what is discussed in the paper!

michael-0brien:
I’ll hold off on pushing more changes until we get a handle on the scaling update, but I’ll get these changes pushed soon after!

optimistix/_solver/levenberg_marquardt.py
michael-0brien: Not sure if you already found this, but there is a [modernization of MINPACK](https://github.com/fortran-lang/minpack) that looks very useful for us! Here is the documentation + source code for the lmder routine, which is based off the implementation from the More paper: https://fortran-lang.github.io/minpack/proc/lmder.html

Looks like the repo is actively maintained too if we want to get in touch.

michael-0brien:
@johannahaffner @patrick-kidger where are we with this? There are some open questions, but as long as we are confident about API and we are sure it is a standard implementation in sciopt, then it seems like we can merge?

I think the only change that would be good before merge is the scaling operator is currently instantiated as `D = diag(-jnp.inf, …)`, but this disallows custom additive update rules like one with momentum, i.e. `D_{k+1} = (1 - p) D_k + (1 - p)**2 D_{k-1} + …` (see above paper; update rules like this are used for determining diagonal SGD preconditioners, which is closely related). I think it should work to instantiate it with zeros instead if the hessian diagonal is positive definite. 

No rush, just a friendly ping + wanted to pass along my updated thoughts :).

pratik-t:
## Hit a bug testing `ScaledLevenbergMarquardt` on `optx.least_squares`

Firstly, thanks for `ScaledLevenbergMarquardt`. I needed it for a fitting problem with parameters spanning several orders of magnitude. However, testing against `optx.least_squares` , `ScaledDampedNewtonDescent.query()` crashes:

`AttributeError: 'ResidualJac' object has no attribute 'hessian'`

I'm not at all familiar with the internals, but this AI suggested fix of branching based on `FunctionInfo` type worked. Please sanity-check the branching logic rather than taking my word for it: 

```python
def query(
    self,
    y: Y,
    f_info: FunctionInfo.EvalGradHessian | FunctionInfo.ResidualJac,
    state: _ScaledDampedNewtonDescentState,
):
    del y

    if isinstance(f_info, FunctionInfo.EvalGradHessian):
        hessian = f_info.hessian
    elif isinstance(f_info, FunctionInfo.ResidualJac):
        hessian = f_info.jac.T @ f_info.jac
    else:
        raise ValueError(
            "ScaledDampedNewtonDescent requires either "
            "EvalGradHessian or ResidualJac."
        )

    scaling_operator = self.update_scaling_fn(hessian, state.scaling_operator)
    return _ScaledDampedNewtonDescentState(f_info, scaling_operator)
```

With this, `ScaledLevenbergMarquardt` converges cleanly on my problem (6 parameters spanning ~O(1) to O(100)) in ~100 steps, where unscaled LevenbergMarquardt on the same problem never terminated within 500 steps.

michael-0brien:
Hi @pratik-t, yes it looks like I did not properly implement the query method, good catch.

I’ve recently implemented a more advanced scaled LM version with the help of AI and I think some changes to this PR are appropriate — some aspects are a bit naive in its current state. But I’m really glad to hear that it’s working well as is!

PR #194, opened 2025-12-10
Drop support for Python 3.10

bagibence:
- Set minimum Python version to 3.11, so future Optimistix versions wouldn't be installable on 3.10. Is that fine?
- In the CI run the tests on 3.11 and 3.12. How about including 3.13 and 3.14?

Follows [JAX's version support policy](https://docs.jax.dev/en/latest/deprecation.html).


bagibence:
Sure! Updated the front pages and the CI runs on 3.11 and 3.13 now.

PR #200, opened 2025-12-24
Integrate benchmarking

johannahaffner:
This follows up on https://github.com/patrick-kidger/optimistix/pull/128. 

Two small things:

- the script stays - having the benchmarks output this would entail customising the fixture provided by `pytest-benchmark`, and plotting is also something one would do every few runs, rather than every run - in particular whenever one would like to look at a comparison between solvers or implementation with others.
- We'll continue to require benchmarking runs to specify `pytest benchmarks --benchmark-only` - if we'd put `--benchmark-only` into `benchmarks/conftest.py` then it would get collected when `pytest` is invoked and it would override our default `--benchmark-skip`. 

And one weird thing: if I remove the `OrderedDict`, I still get results jumbled between solvers. But as you (@patrick-kidger) [note](https://github.com/patrick-kidger/optimistix/pull/128#discussion_r2443247216), it should not be needed. Playing with it for a bit I can remove it in two of the three places, but the underlying logic is not apparent to me, nor am I happy to trust that this would work for a different number of solvers / problems / something. 
I think the pragmatic thing to do here is to simply document the puzzling necessity of `OrderedDict` in this script. (Low-key itching to check whether this is an upstream thing...)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,11 @@ python benchmarks/profile.py <platform> <python_version> <precision> <iD> <kind>
 ```
 
 Where platform refers to the platform on which the benchmarks were run (e.g. Darwin), precision is the numerical precision, e.g. 32bit, and iD is the benchmark run, a four-digit integer.
-These are necessary to identify the saved results for the specific run. `kind` specifies if `runtime` or `compilation` benchmarks are to be compared, and solver names should be given as strings. These are defined in `benchmarks/test_benchmarks.py` for every benchmarked solver, e.g. `optx.BFGS`.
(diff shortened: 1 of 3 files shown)


johannahaffner:
Tests are failing due to JAX 0.8.2 being used with 3.12, and a different version of JAX being used with 3.10, for which JAX has recently dropped support.

johannahaffner:
Now failing checks because `dev` includes a module that triggers the error being addressed in https://github.com/patrick-kidger/lineax/pull/184. Tests pass and benchmarks run locally (benchmarks are skipped in CI and no other tested code is added).

PR #209, opened 2026-02-01
Add Newton-bisection hybrid root-finding method

gerlero:
I've used this method more times than I can count; makes sense to me to suggest an implementation here.

### DIFF
--- a/docs/api/root_find.md
+++ b/docs/api/root_find.md
@@ -41,6 +41,13 @@
 
 ---
 
+::: optimistix.NewtonBisection
+    options:
+        members:
+            - __init__
+
+---
+
 ::: optimistix.BestSoFarRootFinder
     options:
         members:
--- a/docs/how-to-choose.md
+++ b/docs/how-to-choose.md
@@ -26,7 +26,7 @@ If your problem is particularly messy, as discussed above, then use a first-orde
 
 ## Root-finding problems
 
-For one-dimensional problems, use [`optimistix.Bisection`][].
+For one-dimensional problems, use [`optimistix.Bisection`][] or [`optimistix.NewtonBisection`][].
 
 For relatively "well-behaved" problems, then either [`optimistix.Newton`][] or [`optimistix.Chord`][] are recommended.
 
--- a/optimistix/__init__.py
+++ b/optimistix/__init__.py
@@ -71,6 +71,7 @@ from ._solver import (
     LinearTrustRegion as LinearTrustRegion,
     NelderMead as NelderMead,
     Newton as Newton,
+    NewtonBisection as NewtonBisection,
     NewtonDescent as NewtonDescent,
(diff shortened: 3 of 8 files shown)


JiwaniZakir (commented):
In `NewtonBisection.step`, the `error` stored into `new_state` is the function value evaluated at the incoming `y` (before the Newton/bisection update), but `terminate` is then called with the *new* `y` while checking `state.error` — meaning convergence on `f_small` always lags one iteration behind the actual current point. This is inconsistent with using `jnp.abs(y)` as the scale in the same `terminate` call and could result in an unnecessary extra iteration near convergence, or subtly wrong termination behavior if the function has a steep gradient.

There's also a minor docstring issue in the class-level docstring: "the the root" appears twice, and the `__init__.__doc__` string appears truncated in the diff — worth confirming the full doc renders correctly in the mkdocs output.

One edge case worth testing: `deriv == 0` (flat function at the current point). The Newton step `y - error / deriv` produces `±inf`, which correctly falls outside `[new_lower, new_upper]` and triggers the bisection fallback — this is handled gracefully, but it would be worth adding an explicit test case to document and protect this behavior since it's non-obvious.

gerlero:
@patrick-kidger @JiwaniZakir sorry for the delay and thanks for your observations.

> This looks good to me! Can you add this to the existing `Bisection` tests (which seem to be erroneously hanging out in `test_fixed_point.py`, I just now notice...), and other than that I'd be happy to merge this :)

I moved them to  a new `test_bisection.py` file (`test_root_find.py` is obviously also a viable option; I just thought keeping the bisection stuff separate made sense)

> In `NewtonBisection.step`, the `error` stored into `new_state` is the function value evaluated at the incoming `y` (before the Newton/bisection update), but `terminate` is then called with the _new_ `y` while checking `state.error` — meaning convergence on `f_small` always lags one iteration behind the actual current point. This is inconsistent with using `jnp.abs(y)` as the scale in the same `terminate` call and could result in an unnecessary extra iteration near convergence, or subtly wrong termination behavior if the function has a steep gradient.

Good catch. I changed it so that the evaluation of the function is done at the end of the step instead of at the beginning (the derivative is still evaluated at the beginning). Does this way of doing it make sense to you?

> There's also a minor docstring issue in the class-level docstring: "the the root" appears twice, and the `__init__.__doc__` string appears truncated in the diff — worth confirming the full doc renders correctly in the mkdocs output.

Fixed. Also tested building the docs locally and they render fine.

> One edge case worth testing: `deriv == 0` (flat function at the current point). The Newton step `y - error / deriv` produces `±inf`, which correctly falls outside `[new_lower, new_upper]` and triggers the bisection fallback — this is handled gracefully, but it would be worth adding an explicit test case to document and protect this behavior since it's non-obvious.

Agreed. Added such a case.

PR #210, opened 2026-02-07
Added damping factor to fixed point iteration

aidancrilly:
Just a small addition to the fixed point iteration solver to include damping into the Picard iterations. This can help stabilise the method.

### DIFF
--- a/benchmarks/levenberg-marquardt.py
+++ b/benchmarks/levenberg-marquardt.py
@@ -137,11 +137,8 @@ class NormalCholesky(lx.AbstractLinearSolver):
     def conj(self, state, options):
         assert False
 
-    def allow_dependent_columns(self, operator):
-        assert False
-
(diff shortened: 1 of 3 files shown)


benchmarks/levenberg-marquardt.py
```diff
@@ -137,11 +137,8 @@ def transpose(self, state, options):
     def conj(self, state, options):
         assert False
 
-    def allow_dependent_columns(self, operator):
-        assert False
-
-    def allow_dependent_rows(self, operator):
-        assert False
+    def assume_full_rank(self) -> bool:
```
jpbrodrick89, line 140: Can't we just use lx.Normal(Cholesky), now not relevant to this PR just something I noticed when checking in ☺️?

patrick-kidger: Probably yes :D

Issue #182, opened 2025-10-31
Flexibility in how solves are terminated


michael-0brien:
Hello again! First of all, amazing package. The modularity here is really fantastic.

For me, I think a key missing feature is some kind of flexibility in how to terminate solves. There are two aspects of this I think

1) I sometimes struggle with the interpretability of the cauchy termination criteria. My parameters all have different scales, and the meaning of “atol” and “rtol” applied to all parameters is not always meaningful. I would find it useful if there were 1) a way to converge just on function tolerance or 2) to have pytree-valued rtol and atol applied to parameters. 

2) It is very cool that `OptaxMinimiser` is provided. For me, the ability to do the converse—wrap an optimistix minimizer into an optax-like style API—would be very useful. For my work, I don’t always want to use the cauchy termination criteria at all—I may want to just step through a solve and inspect the results. I also may want to vmap over solves, and I don't think it is always wise to use while loop logic in this case. I wrote my own API to give this behavior using `AbstractMinimiser.init` and `AbstractMinimiser.step`, and this was fairly straight forward to do! But I definitely needed the source code to do so and some prior Equinox expertise. Also, related to the above, rtol and atol I noticed were required parameters. 

I realize what I’m describing is not a small feature request by any means and there is no immediate need for this on my end, but I thought it would be interesting to have some discussion about this. Let me know what you think!

michael-0brien:
> 1. I think that different termination criteria could in principle be supported - this is a use case [@bagibence](https://github.com/bagibence) is also interested in, if I recall correctly? This should be fairly simple to do, we could create a callable solver attribute that defaults to `cauchy_termination` and may be overridden by an interested user by providing any other callable with the right signature when instantiating the solver. On our end, this would only require a minor break (adding a solver attribute), and it would give power users a lot more flexibility.

Cool! This would solve things in a pretty straight forward way. Rather than having a pytree-valued rtol and atol, I believe this means I could do appropriate rescalings in this function. 

> 2. For the second request - interactive solving is already supported: https://docs.kidger.site/optimistix/examples/interactive/, as you note. If you have an Optax-style API that works for you, that sounds like a good candidate for the [Awesome List](https://docs.kidger.site/equinox/awesome-list/), exactly the kind of thing it is meant for - a small tool that accomplishes a specific, ecosystem-relevant task.

[...]

PR #214, opened 2026-02-12
Flexible convergence criteria: an `AbstractConvergence` interface

michael-0brien:
Welp, brace yourself: here's an initial attempt for addressing #182. I propose creating an interface for handling terminations:

```python

class AbstractTermination(eqx.Module):

    @abc.abstractmethod
    def __call__(self, y, y_diff, f, f_diff):
        ...

class CauchyTermination(AbstractTermination[Y]):

    rtol: float
    atol: float
    norm: Callable[[PyTree], Scalar] = max_norm

    @override
    def __call__(self, y, y_diff, f, f_diff) -> Bool[Array, ""]:
        ...
```

I've put together a very rough draft with this proposal, where tests are running for minimizers and least squares solvers. Halfway into working on this I realized that `rtol` and `atol` and pretty baked into the code and it's a pretty large refactor. However, I think this would be a significant enhancement to `optimistix` as I don't think that changing termination criteria are edge cases (for example, [`scipy.optimize.least_squares`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html) has different tolerances and norms for function vs. parameter convergence). This is particularly needed for messy problems on noisy scientific data. There are a few current issues implementing this in practice:

- Currently I've clobbered the root finders and the ability to convert problems between one another (e.g. least squares -> minimize or root find -> least squares). This is because I needed to remove the `atol`, `rtol`, and `norm` from the `AbstractIterativeSolver` interface. This could potentially be replaced by the `AbstractTermination` interface, but I didn't want to do this without discussion (also, it is a little out of my depth).
- I've written this as a non-API breaking change, where advanced users of `optimistix` will be able to take advantage of the new interface. It is particularly easy for users to do this for cases where implementing a new solver requires only writing an `__init__` (i.e. the mixing and matching approach), but it is not-so-elegant to take advantage of the new changes for cases where the `AbstractMinimiser`, etc are subclassed directly to a concrete class (e.g. in the case of the `OptaxMinimiser`). In these cases, the user can resort to a tree_at call. This isn’t so bad but would be good to talk about what the right approach is.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -7,7 +7,7 @@ jobs:
   run-test:
     strategy:
       matrix:
-        python-version: [ "3.10", "3.12" ]
+        python-version: [ "3.11", "3.13" ]
         os: [ ubuntu-latest ]
       fail-fast: false
     runs-on: ${{ matrix.os }}
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ examples/data
 .DS_Store
 .devcontainer
 .benchmarks
+uv.lock
+.python-version
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,4 +33,4 @@ repos:
     rev: v1.1.405
     hooks:
     - id: pyright
-      additional_dependencies: ["equinox", "jax", "lineax", "pytest", "optax", "diffrax"]
+      additional_dependencies: ["equinox", "jax", "lineax", "pytest", "optax", "diffrax", "sif2jax", "fire", "matplotlib"]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Contributions (pull requests) are very welcome! Here's how to get started.
 
 ---
 
-**Getting started**
+## Getting started
 
 First fork the library on GitHub.
 
@@ -19,7 +19,7 @@ pre-commit install  # `pre-commit` is installed by `pip` on the previous line
 
 ---
 
-**If you're making changes to the code:**
+### If you're making changes to the code:**
 
 Now make your changes. Make sure to include additional tests if necessary.
 
@@ -30,6 +30,13 @@ pip install -e '.[tests]'
 pytest  # `pytest` is installed by `pip` on the previous line.
 ```
 
+If your changes could affect solver (or compilation) performance, please run the benchmark tests with 
+
+```bash
+pytest benchmarks/ --benchmark-only
+```
+
+You can run benchmarks before or after your change, and also save more extensive results for analysis. For more on this, skip to the "Benchmarking" section below. 
 Then push your changes back to your fork of the repository:
 
 ```bash
@@ -40,7 +47,7 @@ Finally, open a pull request on GitHub!
 
 ---
 
-**If you're making changes to the documentation:**
+### If you're making changes to the documentation:
 
 Make your changes. You can then build the documentation by doing
 
@@ -50,3 +57,44 @@ mkdocs serve
 ```
 
 You can then see your local copy of the documentation by navigating to `localhost:8000` in a web browser.
+
+
+## Benchmarking
+
+If you're interested in more extensive benchmarking - for instance when contributing a new solver - this section is for you. (Note that benchmarks are not run by default, and `--benchmark-only` is required to override this.)
+
+You can save benchmark results with
+
+```
+pytest benchmarks/ --benchmark-save=<benchmark_name> --benchmark-only
+```
+
+and compare against previous runs with `pytest --benchmark-compare`, which will automatically pull in the last saved commit, but also takes run iDs as arguments. See the `pytest-benchmark` [documentation](https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options) for more command line options. 
+The `benchmark-autosave` option will specify the commit iD, instead of a user-defined name.
+Make sure that you are running benchmarks with a clean working tree, so you can trace how changes affect performance!
+
+For convenience, we support some custom flags: 
+
+- `--min-dimension=<int>, --max-dimension=<int>` benchmarks can be run on a subset of problems based on problem size.
+- `--scipy` benchmarks of our solvers are run against the corresponding Python implementation. You might want to limit problem dimension here, they can be quite slow.
+
+pytest's `-k` flags also work in this setting to enable selective execution of benchmarking functions.
+
+**Analysing benchmark results**
+
+You can find a script to analyse benchmark results in `benchmarks/profile.py`. Run it with
+
+```bash
+python benchmarks/profile.py <platform> <python_version> <precision> <iD> <kind> *solver_names
+```
+
+Where platform refers to the platform on which the benchmarks were run (e.g. Darwin), precision is the numerical precision, e.g. 32bit, and iD is the benchmark run, a four-digit integer.
+These are necessary to identify the saved results for the specific run. `kind` specifies if `runtime` or `compilation` benchmarks are to be compared, and solver names should be given as strings. These are defined in `benchmarks/test_benchmarks.py` for every benchmarked solver, e.g. `optx.BFGS`. Putting this together, an example call would be
+
+```bash
+python benchmarks/profile.py Darwin 3.13 64bit 0001 runtime optx.BFGS optx.LBFGS
+```
+
+**If you are contributing a solver**
+
+In this case, you're probably reasonably familiar with the alternatives out there - if implementations we could compare to exist, please add them to the listed solvers in `benchmarks/test_benchmarks.py`, including hyperparameters such as solver tolerances to get as fair of a comparison as is feasible.
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Features include:
 pip install optimistix
 ```
 
-Requires Python 3.10+ and JAX 0.4.38+ and [Equinox](https://github.com/patrick-kidger/equinox) 0.11.11+.
+Requires Python 3.11+.
 
 ## Documentation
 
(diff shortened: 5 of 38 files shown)


michael-0brien:
@jpbrodrick89 this sounds reasonable to me! If I had to guess I think this PR is probably most likely to be pushed through if a comprehensive alternative is offered compared to the current model for terminations, so extending this to root finders sounds great. I could use the help as I don’t work on root finders. If you also want to take a whack at fixing where root finds are converted to minimizations and least squares solves please go for it as well!

michael-0brien:
@patrick-kidger @johannahaffner when you have a moment could you help me come to a solution on this? I can work on this but want to make sure it’s the right direction, and if not what would be. There are two approaches to resolution in my mind:

1. Something like the current approach (add an API).
2. Some solution where flexible termination can be solved by overwriting `terminate` methods (keep API the same but allow for hacking). While this already works for some solvers, it does not for others like  `AbstractQuasiNewton` classes

I would lean towards 1. because at the least optimistix should support the ability to have different norms and tolerances for y and f as this is quickly broken by many problems, but curious if there is a feasible path for 2. as well.

michael-0brien:
To present an alternative to the current version of things, it would be possible to preserve `atol` and `rtol` as a part of the `AbstractIterativeSolver` API. This could look like the following:

```python
class AbstractTerminationCriteria(eqx.Module):

    atol: eqx.AbstractVar[float]  # these are forwarded to the `AbstractIterativeSolver` API, when appropriate
    rtol: eqx.AbstractVar[float]
    norm: eqx.AbstractVar[Callable]

    @abc.abstractmethod
    def __call__(...):
        ...

class CustomTerminationCriteria1(AbstractTerminationCriteria):

    atol: float  # atol and rtol are used for function tolerance
    rtol: float
    yatol: PyTree[float]  # pytree-valued atol and rtol for parameter tolerance 
    yrtol: PyTree[float]
    norm: Callable

    def __call__(...):
        ...

class CustomTerminationCriteria2(AbstractTerminationCriteria):

    atol: float
    rtol: float
    norm: Callable

    def __call__(...):
        #  atol and rtol only used for function tolerance; no parameter tolerance considered
        ...
 
```

Under this idea I am still not sure how to handle the issue of problem conversion (i.e. using a minimiser to solve a least squares problem). In either case this is the only thing that breaks the proposed API. I personally think this is a bit messier, but it is less API breaking in case users are accessing `solver.atol/rtol/norm`.

jpbrodrick89:
Apologies for the lack of update on my side, the reason I've not attempted this for root finders yet is we're still trying to determine the correct non-Cauchy termination conditions for Newton.

michael-0brien:
@jpbrodrick89 no problem! There's no rush, I mostly want to discuss direction so I can start using this PR in my own fork. I also quickly want to correct myself:

> Under this idea I am still not sure how to handle the issue of problem conversion (i.e. using a minimiser to solve a least squares problem). In either case this is the only thing that breaks the proposed API. I personally think this is a bit messier, but it is less API breaking in case users are accessing `solver.atol/rtol/norm`.

Looking through the code a little bit more closely, actually root finder problem conversion (i.e. using least squares solver or a minimiser to solve a root find) is the only case where this PR breaks. This is good news actually! I bet we can find a simple way to address this. Further, the current method of handling terminations here looks slightly hacky so I bet there's an elegant solution.

johannahaffner:
Sorry for the slow replies, @michael-0brien! I was mulling this over a few days ago, and indeed we should also think about @jpbrodrick89's change to the termination criteria for the root finders in this context. 

I'll try to sit down with pen and paper in the coming days and map this out. 

I prefer an API break with a clean solution over some hybrid, and since concrete solvers already have an `__init__`, forwarding to wherever we think is appropriate is easy. The fraction of power users that define solver attributes for custom solvers is, I think, competent enough to adjust their code and upgrade to declaring new attributes.

michael-0brien:
Thanks so much @johannahaffner! Let me know how I can help—maybe there is a way to adapt the current PR to encompass changes to the root finder behavior. This would make things much more complete, the current draft really only focuses minimizers and lstsq solvers!

One argument for something like the current proposal, which I haven’t had a chance to write a draft of, is the `AbstractTermination` API may fit nicely into the advanced API docs that currently has searches and descents. Thematically it is very similar—there is default behavior for every optimizer in optimistix, but it is extremely straightforward to create a concrete class with custom searches, descents, and now possibly termination criteria as well.

michael-0brien:
Hi, ping! Wanted to check in. Wanted to share how I happen to be using this interface downstream:

```python

class CustomTermination(optx.AbstractTermination[Y]):
    ftol: float | tuple[float, float]
    ytol: float
    y_scale: float | PyTree[float]
    norm: Callable = optx.max_norm

    @override
    def __call__(self, y: Y, y_diff: Y, f: F, f_diff: F) -> Bool[Array, ""]:
        eps = jnp.finfo(jax.dtypes.canonicalize_dtype(float)).eps
        if isinstance(self.ftol, Sequence):
            assert len(self.ftol) == 2
            frtol, fatol = self.ftol
        else:
            frtol, fatol = self.ftol, eps
        y_scale, f_scale = self.y_scale, ω(f).call(jnp.abs).ω
        if isinstance(y_scale, float):
            y_scale = _tree_full_like(y_diff, y_scale)
        y_converged = (
            self.norm((ω(y_diff).call(jnp.abs) / (eps + self.ytol * y_scale**ω)).ω) < 1
        )
        f_converged = (
            self.norm((ω(f_diff).call(jnp.abs) / (fatol + frtol * f_scale**ω)).ω) < 1
        )
        return y_converged & f_converged
```

This a) separates ytol and ftol criteria, b) allows per-parameter convergence criteria via a PyTree-valued `y_scale`, and c) overrides `optimistix`'s default behavior of `y_scale = (ω(y).call(jnp.abs)).ω`, which for some optimization problems, like what I am doing, would be a bug / is an incorrect thing to do.

Let me know if anyone has an updated thoughts on this PR (@patrick-kidger & @johannahaffner?). Thanks so much in advance and no rush, I understand how busy the maintainers must be.

patrick-kidger:
'Busy' is the right word these days... @johannahaffner and I recently made the questionably-wise decision of founding a startup :D This does mean that we're picking-and-choosing our battles a little more carefully on which PRs to work through and accept.

For that reason then right now I think – realistically – this PR is probably one of the lower-priority ones. Termination can anyway be overriden through subclasing/composing, so IIUC this is really a bit of polish on what is already a fairly advanced feature? Please let me know if you think I've misunderstood something there.

michael-0brien:
> 'Busy' is the right word these days... @johannahaffner and I recently made the questionably-wise decision of founding a startup :D This does mean that we're picking-and-choosing our battles a little more carefully on which PRs to work through and accept.

Makes sense!

> For that reason then right now I think – realistically – this PR is probably one of the lower-priority ones. Termination can anyway be overriden through subclasing/composing, so IIUC this is really a bit of polish on what is already a fairly advanced feature? Please let me know if you think I've misunderstood something there.

This is true in some cases, but not all. E.g. in `AbstractBFGS` termination cannot be overwritten without overwriting the hessian update. I am forgetting right now if this is an exception or if there are other cases as well, but I recall there are other cases as well where the termination method ends up simply returning a boolean passed from the state pytree.

I would also argue that there is a reason frameworks like scipy.optimize have more built in control of termination (i.e. scipy.optimize.minimize). There it is top-level API, not an advanced feature; my guess being that it is an important part of making the algorithms widely applicable. In making optimistix the scipy.optimize of JAX, I think this issue while small is pretty significant. Others I have spoken to also are in the world of “inference of physical parameters on experimental data in the presence of noise” seem to have similar experience.

Here’s an example I have not yet already mentioned in this thread: consider optimizing the translation of an object; i.e. there is noisy data with an object in frame, and you want to find how to shift a model to the reference. Let’s say the domain is [-100, 100], for the sake of demonstration. The object can be anywhere in frame, therefore whether it is located at x = 90.0004 or x = 0.0004, say, my localization precision does not change in terms of absolute x. Let’s say my noise threshold is such that I can only get three decimals of precision. In either case my x_scale should be ~ 10^-3, where my optimizer loses sensitivity. What is currently rtol when applied to x does not have meaning.

In this toy example, one could mess around with rtol and try to only converge on atol. But introduce two parameters, each with their own properties and sensitivities, and optimistix can’t accommodate. In any case, this is just one aspect added onto previous discussion in this thread; perhaps the biggest limitation in practice is still having the same rtol and atol for function and parameter convergence.

I understand things are quite busy and thank you for making the time to respond! I do feel passionately that this issue is an important one, and I suppose the only reason not to include it is that it is slightly non-trivial to engineer into the current code. Another route would be to consider making it truly possible to factor out each terminate method in the library, but I think I may have found that more complicated when I wrote this PR.

michael-0brien:
Hi all, I've resolved outstanding issues and gotten this PR to the point of review. There has been a lot of discussion in this thread (mostly from me, haha), but to TL;DR:

1) To factor out termination criteria, we cannot easily overwrite `terminate`. As one example, `AbstractQuasiNewton` stores the convergence check in the state and passes it through `terminate`.
2) Therefore, I have proposed that we add an interface for the convergence check called `AbstractConvergence` with a single method called `check`.
3) To achieve this, it is necessary to remove `atol`, `rtol`, and `norm` on the `AbstractIterativeSolver` interface. Aside from adding this feature, I think this was an over-constrained design for the library and is generally a healthy thing to do.

Backwards compatibility: `AbstractRootFinder` and `AbstractFixedPointSolver` are fully backwards compatible as I have added `atol`, `rtol`, and `norm` to their interfaces. If there are changes to this in the future, I think this is a separate PR. `AbstractMinimiser` and `AbstractLeastSquaresSolver` lose their `atol`, `rtol`, and `norm` pytree leaves and gain a `convergence` abstract field. These could be declared as properties for more backwards compatibility. 

I would have liked the `convergence` abstract field to be declared at the same level of abstraction as searches and descents (e.g. in `AbstractQuasiNewton`, rather than `AbstractMinimiser`). It was necessary to have the new interface on `AbstractMinimiser` to preserve the problem-conversion behavior in _root_find.py.

@patrick-kidger @johannahaffner @jpbrodrick89 no rush but when you get a moment, please do take a good look and let me know if you have any questions!

PR #217, opened 2026-02-16
Anderson acceleration for fixed point iteration

aidancrilly:
Continued of closed #215 - apologies if this is not the way to re-open

I have implemented a form of Anderson acceleration for fixed point problems which should increase the rate of convergence in fixed point problems over standard fixed point iteration. Mentioned in #3 

Here is a simple example for cos(x) = x:
image

A few queries:
- I had to run the tests at very low tolerance for the finite difference jvp based tests to pass on the new solver. Through some separate investigation this seems to be from the small epilson used and the potentally non smooth behaviour of solver when you allow larger tolerances. Happy to investigate further if this is of concern.
- The default fixed point iteration solver appears to work for complex numbers but I doubt the Anderson acceleration would (not something I know about). I assume it is fine to mark it as not compatible with complex and skip these tests as I have done?
- I needed a helper routine (_batched_tree_zeros_like) very similar to one in LBFGS but reversed shape. Don't know if this is confusing or if more generic helpers would be of use.

### DIFF
--- a/optimistix/_solver/fixed_point.py
+++ b/optimistix/_solver/fixed_point.py
@@ -24,9 +24,8 @@ class FixedPointIteration(AbstractFixedPointSolver[Y, Aux, _FixedPointState]):
     This is one of the simplest ways to find a fixed point `y` of `f`: simply
     repeatedly call `y_{n+1}=f(y_n)` until `y_n` stops changing.
 
-    Optionally, one can use damping between iterations, then the update becomes:
-    `y_{n+1}=(1-damp)*f(y_n)+damp*y_n`
-    Where damp is the damping factor.
+    Optionally, one can use damping between iterations, then the update becomes
+    `y_{n+1}=(1-damp)*f(y_n)+damp*y_n` where `damp` is the damping factor.
 
     Note that this is often not a very effective method, and root-finding algorithms are
     frequently preferred in practice.
@@ -60,7 +59,9 @@ class FixedPointIteration(AbstractFixedPointSolver[Y, Aux, _FixedPointState]):
         tags: frozenset[object],
     ) -> tuple[Y, _FixedPointState, Aux]:
         fn_y_n, aux = fn(y, args)
-        new_y = jtu.tree_map(lambda i_np1, i_n : (1-self.damp)*i_np1+self.damp*i_n, fn_y_n ,y)
+        new_y = jtu.tree_map(
+            lambda i_np1, i_n: (1 - self.damp) * i_np1 + self.damp * i_n, fn_y_n, y
+        )
         error = (y**ω - new_y**ω).ω
         with jax.numpy_dtype_promotion("standard"):
             scale = (self.atol + self.rtol * ω(new_y).call(jnp.abs)).ω
--- a/optimistix/_solver/fixed_point.py
+++ b/optimistix/_solver/fixed_point.py
@@ -1,10 +1,11 @@
 from collections.abc import Callable
-from typing import Any
+from typing import Any, Generic
 
 import equinox as eqx
 import jax
 import jax.numpy as jnp
 import jax.tree_util as jtu
+import lineax as lx
 from equinox.internal import ω
 from jaxtyping import Array, Bool, PyTree, Scalar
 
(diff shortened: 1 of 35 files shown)


optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +94,217 @@ def postprocess(
...
+        tags: frozenset[object],
+    ) -> _AndersonAccelerationState:
+        del fn, args, options, f_struct, aux_struct
+        state = _AndersonAccelerationState(
+            first_step=jnp.array(True),
+            index_start=jnp.array(0),
+            history_length=self.history_length,
+            f_history=_batched_tree_zeros_like(y, self.history_length),
+            g_history=_batched_tree_zeros_like(y, self.history_length),
+            output_structure=jax.eval_shape(lambda: y),
```
patrick-kidger: I don't think `output_structure` needs to be in the state, it can be constructed in `step` where it is used.

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +94,217 @@ def postprocess(
...
+    relative_error: Scalar
+
+
+class AndersonAcceleration(
+    AbstractFixedPointSolver[Y, Aux, _AndersonAccelerationState]
+):
+    """
+    Solves the fixed-point equation
+
+        x = f(x)
```
patrick-kidger: Nit: we call this quantity `y`, not `x`, in Optx.

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +94,217 @@ def postprocess(
...
+    where Δf_k contains differences of recent mapped iterates f(x).
+
+    A damping factor `damp ∈ [0, 1]` optionally blends the accelerated
+    update with the previous iterate:
+
+        x_{k+1} ← (1 - damp) * x_{k+1} + damp * x_k.
+
+    Reference:
+    D. G. Anderson. Iterative procedures for nonlinear integral equations. J. Assoc.
+    Comput. Machinery, 12:547–560, 1965.
```
patrick-kidger: Nit, I try to refer to references with a bibtex reference, for ease of copy-paste. (Do a grep for `bibtex` and you'll see a couple of examples for reference.)

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +94,217 @@ def postprocess(
...
+            lambda x, z: x.at[..., state.index_start % self.history_length].set(z),
+            state.g_history,
+            g_y_n,
+        )
+
+        # Perform Anderson acceleration
+        # First iterate must be treated differently
+        def _first_iterate():
+            new_y = jtu.tree_map(
+                lambda y_old, y_next: self.damp * y_old + (1 - self.damp) * y_next,
```
patrick-kidger: Nit: `if isinstance(self.damp, (int, float)) and self.damp == 0` – which is the fairly common default, and is something known at compile time – then we can skip this operation and probably save some effort.

This is the kind of thing that XLA will typically not compile out because in principle `0 * nan = nan`, so it can't apply the 'mathematical' optimization that `0 * x = 0`.

patrick-kidger (commented):
Okay, this looks really clean to me! I have just some minor nits.

Can you also add this to the documentation?

optimistix/_solver/fixed_point.py
aidancrilly: Ah interesting! I did not know it might not be compiled out, made the change here (as well as in the damped fixed point iteration scheme).

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+    Take rolling buffer tree and compute diff between iterations, and create operator
+    """
+
+    indices = jnp.maximum(i - jnp.arange(0, batch_dimension - 1), 0) % batch_dimension
+    shift_indices = jnp.maximum(i - jnp.arange(1, batch_dimension), 0) % batch_dimension
+    diff_tree = jtu.tree_map(
+        lambda leaf, shift_leaf: leaf[..., indices] - shift_leaf[..., shift_indices],
+        tree,
+        tree,
+    )
```
patrick-kidger, line 118: It looks like you're passing `tree` as the argument twice. I think just
```python
jtu.tree_map(lambda x: x[..., indices] - x[..., shift_indices], tree)
```
should work here?

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+            year={1965},
+            publisher={ACM New York, NY, USA}
+        }
+        ```
+    """
+
+    rtol: float
+    atol: float
+    history_length: int = 10
+    norm: Callable[[PyTree], Scalar] = max_norm
```
patrick-kidger, line 189: Ultranit: I try to keep `rtol`/`atol`/`norm` adjacent and in that order (which matters when initializing the solver with positional arguments).

On a more important note, do you think it's worth setting a default value for `history_length` or should we expect a user to suggest a value? I don't know AA well enough to know if this value is a common choice across many problems.

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+        options: dict[str, Any],
+        f_struct: PyTree[jax.ShapeDtypeStruct],
+        aux_struct: PyTree[jax.ShapeDtypeStruct],
+        tags: frozenset[object],
+    ) -> _AndersonAccelerationState:
+        del fn, args, options, f_struct, aux_struct
+        state = _AndersonAccelerationState(
+            first_step=jnp.array(True),
+            index_start=jnp.array(0),
+            history_length=self.history_length,
```
patrick-kidger: I don't think `self.history_length` needs storing in the state, it can still just be accessed as `self.history_length`.

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+        state: _AndersonAccelerationState,
+        tags: frozenset[object],
+    ) -> tuple[Y, _AndersonAccelerationState, Aux]:
+        fn_y_n, aux = fn(y, args)
+        g_y_n = jtu.tree_map(lambda a, b: a - b, fn_y_n, y)
+        output_structure = jax.eval_shape(lambda: y)
+
+        # Add new results to history
+        f_history = jtu.tree_map(
+            lambda x, z: x.at[..., state.index_start % self.history_length].set(z),
```
patrick-kidger: Normally I'd expect that we want to indexing into the first dimension so that we're writing `z` into a contiguous block of memory. I think it'd be worth moving the history dimension to the start rather than the end.

I'm aware that XLA will reorder dimensions under-the-hood, but I'm not sure how much I really trust it in practice.

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+            # Compute gammas
+            # Use SVD - for initial iterations, F has columns of zeros
+            gammas = lx.linear_solve(G_op, g_y_n, solver=lx.SVD()).value
+
+            # Compute terms for update and include damping
+            F_gamma = F_op.mv(gammas)
+
+            new_y_undamped = jtu.tree_map(
+                lambda f_n, F_gamma: f_n - F_gamma, fn_y_n, F_gamma
+            )
```
patrick-kidger: FYI we have a `from equinox.internal import ω` that can be used to rewrite this (and other similar tree-map'd equations) as `(fn_y_n**ω - F_gmma**ω).ω` It doesn't handle all possible tree-map'd operations but you'll see it a lot in the Diffrax/Optimistix/etc codebases to help make this kind of thing more readable.

(At the expense of needing to be able to type a greek letter on your keyboard, of course... :p This being a choice I made at a time when I was still firmly a full-time mathematician!)

optimistix/_solver/fixed_point.py
```diff
@@ -93,6 +101,240 @@ def postprocess(
...
+        tree,
+    )
+
+    pytree_op = lx.PyTreeLinearOperator(diff_tree, output_structure)
+    return pytree_op
+
+
+class _AndersonAccelerationState(eqx.Module, Generic[Y]):
+    first_step: Bool[Array, ""]
+    index_start: Scalar
```
patrick-kidger: I don't really mind, but why `index_start` and not just `index`? I'm not grasping the significance here.

patrick-kidger (commented):
Okay, next revivew! I have only nitty minor comments, this basically looks good to me. :)

optimistix/_solver/fixed_point.py
patrick-kidger: Thank you!

optimistix/_solver/fixed_point.py
aidancrilly: I have set the default value to 5 to be in line with scipy and jaxopt Anderson acceleration implementations

optimistix/_solver/fixed_point.py
aidancrilly: So I stuck with this order since otherwise the pytree operator made from diff_tree was not compatible with output_structure based on y (as gammas needs to history_length long so operator is shape (*y_shape, history_length) ). Perhaps there is a way around this?

unalmis:
Maybe also consider #231 in documentation for this

jlperla:
This looks great. Anything outsiders can do to help get this ready to merge?

PR #227, opened 2026-04-09
refactor line-search solvers to return latest iterate

jpbrodrick89:
As suggested in your [comment](https://github.com/patrick-kidger/optimistix/pull/225#issuecomment-4126518747) in #225, I have refactored line-search solvers, specifically `AbstractGradientDescent`, `AbstractQuasiNewton` and `AbstractGaussNewton` (and their subclasses implicitly), such that `step` always receives and returns the latest iterate rather than the last accepted step. The last accepted step is now stored in `state.y_accepted` instead of `state.y_eval`. This means we don't have to use `tree_where` to construct `prev_aux` anymore and all minimisers and least square solvers compose with `BestSoFarMinimiser` in the same way—including the very latest iterate in the sequence rather than the lagged accepted. Note that this means that the `y` returned will be worse than `y_accepted` on rejected steps, but as `BestSoFarMinimiser` sees all iterates this is not an issue.

Once this is merged I can have a think about whether a wider unification of line-search solvers is possible.

### DIFF
--- a/optimistix/_solver/gauss_newton.py
+++ b/optimistix/_solver/gauss_newton.py
@@ -18,7 +18,6 @@ from .._misc import (
     max_norm,
     sum_squares,
     tree_full_like,
-    tree_where,
 )
 from .._search import (
     AbstractDescent,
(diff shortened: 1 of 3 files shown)


johannahaffner:
Edit: just changed the base to a new `dev` branch.

PR #230, opened 2026-05-12
fix reporting nonconvergence as divergence for non-Cauchy termination

jpbrodrick89:
This would fix #221  if VeryChord subclassed AbstractNewtonChord but it doesn't, as such I will need to mirror this in a diffrax PR. Note the step counts only reduce from 845 to 242 rather than 238 simply because `at_least_two` is now doing what it shold.


patrick-kidger:
Merged! Thank you for identifying what was going wrong, and for the fix. This one is important enough that I'll probably do a new release soon.

PR #234, opened 2026-07-07
Evaluate f_info for real in Gauss-Newton init instead of zero-filling

### LINKED ISSUE #229: NaNs encountered when taking the quotient of two arrays
Hi, I'm quite new to Optimistix (and JAX more broadly), so please forgive the simple question if the solution turns out to be obvious.

I'm attempting to solve a root finding problem which relies quite heavily on external function calls in the residual. These obviously require special treatment, but doesn't seem to actually be the cause of my problems in this instance. The key to this it seems is that taking the quotient of two arrays is producing NaN values in the JVP which propagate through the rest of my code.

```python
import jax

jax.config.update("jax_disable_jit", True)
jax.config.update("jax_debug_nans", True)

class Solver(eqx.Module):
    arr: Float[Array, "n"]

    def residuals(self, y, args):
        z = self._quotient(y)
        return jnp.array([0.5, 0.5]) - z

    def solve(self):
        y0 = jnp.array([0.9, 0.1])
        solver = optx.LevenbergMarquardt(rtol=1e-5, atol=1e-5)

        sol = optx.root_find(
            self.residuals,
            solver,
            y0=y0,
            args=(),
        )

        return sol

    def _quotient(self, y):
        # FloatingPointError: invalid value (nan) encountered in div
        return y / self.arr


if __name__ == "__main__":
    problem = Solver(jnp.array([5.0, 2.5]))
    sol = problem.solve()
    print(sol.value)
```

For this toy example I can re-enable the JIT or disable NaN checking and everything works just fine. However, it seems that the NaN values persist, only that JAX is able to handle them appropriately. The problem still remains for me however, as these NaN values in the tangent will propagate through the pure callbacks and land in my external functions which do not know to handle them appropriately.

I am not really sure if this is a JAX or Optimistix (or more likely my) problem, but given that I can get a Jacobian out of the residual just fine I thought it would be best to start here.

Thanks for any help!

Sumu004:
Thanks @NiallOswald for the root-cause tracing in #229 — this implements what you diagnosed.

`AbstractGaussNewton.init` filled `state.f_info` with zeros via `tree_full_like`, which also zeroes any data closed over inside `f_info.jac` (the `jax.linearize` capture). `ClassicalTrustRegion.predict_reduction` calls `f_info.jac.mv(...)` unconditionally on every step, including the first, so a residual function that's nan-sensitive at a zeroed closure value produces a real `nan` that contaminates the JVP — even though the accept/reject decision itself is separately forced correct on the first step. `GaussNewton` (no trust region) is unaffected; `LevenbergMarquardt`/`InverseLevenbergMarquardt` both inherit the same `init` and are.

Fix: call `_make_f_info` directly in `init` instead of faking its shape with zeros — exactly what `step` already does every other iteration. Costs one extra linearization at init.

I don't have a working JAX install, so I've verified this by tracing the exact call paths rather than running it — would appreciate a maintainer confirming against the real reproducer.

Fixes #229


patrick-kidger:
Thank you for the fix! (And gosh, sorry for the delay getting to this - both your fix and the original issue.)

So I think this might still actually have a slight bug: we'd never update the linearisation point from the initial value used.

This is because of these lines: 

https://github.com/patrick-kidger/optimistix/blob/6415ab15480ebdb7386e17850105ce0dba6786ec/optimistix/_solver/gauss_newton.py#L254-L259

which always keep the "old" jaxpr (as noted in the comment!) with the mistake being it does not distinguish between a `Jaxpr` and a `ClosedJaxpr`. Thus it preserves the entire `ClosedJaxpr`, including the constants.

I've not tracked down exactly what is holding the `ClosedJaxpr` - probably the correct fix is to either update it to use a `Jaxpr` instead, or to adjust the location to also handle `ClosedJaxpr`s (or for robustness, both!)

WDYT?

PR #237, opened 2026-08-10
Fix GoldenSearch initial midpoint computation

jankiemm:
The initial `middle` was computed as an offset from `lower` but never added to `lower`, so it was treated as an absolute coordinate within [lower, upper] everywhere else in the solver. This only happened to work for brackets straddling zero; for others (e.g. [99, 101]) the initial midpoint fell outside the bracket and the solver diverged.


patrick-kidger:
Merged! 🎉 Thank you for the fix!

